PR #1, opened 2010-09-06
fixed some warnings

ghost:
I cleaned up some warnings of comparing signed with unsigned values, and removed some asserts that could never fail (ie: comparing if an unsigned int >= 0, which is always true).

can cherry pick commit 338b55ab51d645d8af95


ricardoquesada:
thanks. Merged.

PR #7, opened 2010-10-15
updated FSAA code

araker:
@Riq, I've seen your comment in the FSAA thread (http://www.cocos2d-iphone.org/forum/topic/7552), but the code from djdrzzy is outdated, the improved FSAA code is inside my last commit.


ricardoquesada:
@araker: thanks. Could you tell me which part is missing ? I didn't use your commit, because your commit modifies many files which I believe are not part of this feature. If I'm wrong, please let me know.

Also, after applying the MSAA patch, I applied a another feature to EAGLView (sharegroup).
So, if possible could you:
- download the latest cocos2d develop version
- fix or update the MSAA code
- and send me a pull request

Thank you for you time!

araker:
The many files in the previous commit were changes to the tests to work with the new EAGLView initializer, but I saw that you've already changed the tests to work with the new EAGLView. So those weren't necessary anymore.

I'll send you a pull request with the updated EAGLView and in the future I'll try to keep my repository up to date before sending any pull requests.

ricardoquesada:
Thank you araker.

PR #11, opened 2010-10-16
fixed wrong assert in CCRenderTexture

araker:
Wrong assert addressed as well as saving the clear color every time the rendering begins, instead of a one time save during initialization.


ricardoquesada:
fixed .thanks.

PR #12, opened 2010-10-17
RenderTexture, small bugfix and added save as raw data option

araker:
Fixed wrong assert
Added save as raw data option, more info at http://www.cocos2d-iphone.org/forum/topic/10178#post-58560


ricardoquesada:
Thanks. Fixed.

PR #13, opened 2010-10-24
CocosDenshion changes

ghost:
Sorry for the delay in getting back to you Riq.  These changes could go in the release candidate, they are just new features but pretty minor so there is no rush.  Steve


ricardoquesada:
Thank for the info. I found a sort of critical bug in  "develop" branch so probably I'll release -rc tomorrow, after fixing this bug.

PR #14, opened 2010-10-26
New particle system type, relative positioning

jordo:
http://www.cocos2d-iphone.org/forum/topic/10622

I use this quite often..  I use this for 'trails' of particles...

You can't use kCCPositionTypeFree for 'trails' or particles if the particle system is attached to a node that moves..  kCCPositionTypeFree always converts it's position to world coordinates..   My game is a side scroller which scrolls a layer..  My particle system is always added as a node on that layer, so in order to get particle trails to work properly I always have to use this type of system.


ricardoquesada:
thanks. merged.

PR #16, opened 2010-11-02
Work around for OS audio session bug

ghost:
This is a work around for an OS bug that has manifested itself in various ways on 4.x, causes loss of OpenAL audio i.e. sound effects.  Latest incarnation occurs when locking the device on 4.2.  Just take CDAudioManager and CocosDenshion files.


ricardoquesada:
thanks. merged.

PR #20, opened 2010-11-17
adds support for chipmunk on mac also

rolandoam:
well, support has always been there, just fixed the equivalency of CGPoint and cpVect (there's a minor header problem there).


rolandoam:
sure!

PR #21, opened 2010-11-29
Reorder child and addChild optimizations

araker:
Reorderchild and insert child optimizations

Now for every added element and reordered element a linear search is performed and items in the children and/or descendants array are shifted when the new/reorderd child is added.

This implementation batches all the changes and performs before drawing one insertion sort on the children array. In case of CCSpriteBatchNode it also gives each child a new atlasIndex and performs one insertion sort on the descendants array.

What has changed?
- CCNode, addChild, reorderChild are changed
  sortAllChildren is added
  isReorderChildDirty_ is added
  visit is slightly changed
- CCSprite
  sortAllChildren is overriden
  reorderChild and addChild are changed
  setReorderChildDirtyRecursively is added
- CCSpriteBatchNode
- addChild and reorderChild are changed
- appendChild, updateAtlasIndexes, sortDescendants are added
- visit is slightly changed, sortAllChildren is changed

-SpriteTest
spriteBatchNodeReorder is changed to make it work with the new sorting method.

What does it do?
- CCNode, a added child is appened to the children array, a reorderd child z-order is set immediately. Sorting happens before drawing. A dirty bool keeps track if sorting is necessary or not.
- CCSprite, overrides sortChildren, when it is added to a batchNode it also sorts its children recursively.

-CCSpriteBatchNode, a new child is appended to the descendants array as well. Sorting is 3 stages. First recursively all children are sorted based on their zOrder. Second every child gets a new atlasIndex, based on their relative zOrder (keeping parent -> child relations intact). Third all descendants and quads are sorted based on the atlasIndex.

All intense looping methods are optimized with fast dispatch.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,4 @@
-version 0.99.5-rc1 - 12-Nov-2010
+version 0.99.5-rc1 - 15-Nov-2010
 . [NEW] Director: setProjection calls "updateProjection" delegate if "custom" projection is used
 . [NEW] Label: The CCLabelProtocol implements the -(NSString*)string method.
 . [NEW] Particles: Added a new type kCCPositionTypeRelative. Useful for "trail" particles. Added Test.
@@ -15,6 +15,7 @@ version 0.99.5-rc1 - 12-Nov-2010
 . [FIX] EAGLView: Doesn't crash in multithread environment (re-fixed issue #350)
 . [FIX] Label: BMFontLabel simplified loading of atlas. Doesn't use CCFileUtils (issue #1031)
 . [FIX] Label: setString copies the string, and not retain it (issue #1034)
+. [FIX] Mac: Events are dispatched when CC_DIRECTOR_MAC_USE_DISPLAY_LINK_THREAD is used
 . [FIX] Menu: MenuItems can draw its children (issue #458)
 . [FIX] Menu: MenuItems positions can be controlled manually (issue #1013)
 . [FIX] Menu: Menus uses touch priority kCCMenuTouchPriority (-128) (issue #1033)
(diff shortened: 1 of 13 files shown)


araker:
Forgot to mention that all spriteTests behave like expected with this code.

araker:
Also if this gets accepted, it would make insertChild, atlasIndexForZ, lowestAtlasIndexInChild and higestAtlasIndexInChild in CCSpriteBatchNode redundant, since they aren't used anymore.

ricardoquesada:
Very Interesting stuff. Have you done any performance tests ? How fast it now ? Thanks.

araker:
Just tried it out and addChild, removeChild perform the same, reorderChild is around 5x to 20x faster. The results of my reorder test were a bit weird, it was initially 5x faster, but it got faster and faster and finally stabilized at the 20x faster level. I tested on a 3GS, release build. 

My github repo is a mess, so I can't send you a pull request with my altered performance test, sorry for that. I've send you a mail instead.

araker:
I was a bit surprised initially that the addChild performance hasn't improved, but now I realize this is the worst case scenario for insertion sort. Insertion sort works particular well when adding elements to a already sorted collection. It can be n log n in best case scenarios. In worst case scenerios, like having to sort a completely random collection, it behaves like n ^ 2. So even in its worst case performance it still has the same performance as the current addChild implementation. In less extreme situations it will probably perform better than the current implementation.

ricardoquesada:
I'll test your patch in a different branch. I'll do some tests with it, and if everything works as expected I'll merge it after v0.99.5

quick update: I did some tests, and reorder is not working properly.

ricardoquesada:
Branch: reorderSprite
https://github.com/cocos2d/cocos2d-iphone/tree/reorderSprite

araker:
Thanks for the link. Which tests did you run to determine that reorder wasn't working properly?

ricardoquesada:
A test in SpriteTest.m called "SpriteBatchNode: reorder issue #766" is not working correctly.

araker:
Ok, I'll look into it somewhere in the next couple of days.

ricardoquesada:
ok. thanks.

araker:
I've corrected the code to make the test cases work, but there is a underlying issue that I still have to correct.

The current implementation always adds or reorders a child with a already present zOrder at the end of that zOrder. 

Eg.   

child      0 1 2 3 
zOrder   2 4 4 4

If you give child 0 a zOrder of 4 it will be placed at spot 3 in the children array. 

The insertion sort on the other hand won't move child 0 since the order is already sorted (it moves childs from bottom to top till it finds a equal or higher number). 

So when using an insertion sort and multiple children have to be reordered to a existing zOrder, the order is 
determined by the place in the array and not by the order of calling reorderChild.

I didn't take the sequence of adding/reordering into account, so I'll add some logic to keep track of that as well.

If there are any other gotcha's with the zOrder, let me know ;)

ricardoquesada:
Thanks!
I'll let you know if I find any other issue.

joekim:
I'll be excited to see this in 0.99.5 if possible.

PR #23, opened 2010-12-03
Chase camera.

twyatt:
Minor updates to allow for custom cameras.

Implemented a simple chase camera as an example use.

Thanks for all your hard work on cocos2d-iphone!

Hope this proves to be a useful contribution.


twyatt:
I understand that CCFollow provides a similar functionality, I just thought that being able to set an alternate camera for a CCNode might be useful and may provide a slightly more intuitive interface for camera modifications?

twyatt:
I'm starting to second guess this camera implementation: I see the clear advantage of having the camera be action based (as you have already implemented), as that provides a delta time to camera actions such as the orbit camera.

So setting a camera directly to a node (as I have implemented in this pull request) might be more intuitive for a chase camera but would become cumbersome for an orbit camera that may need to rely on a delta time (to remain in-sync with a game's step).

Well, I suppose just let me know your opinion on the subject.

And once again, thanks for cocos2d-iphone!

PR #29, opened 2011-01-25
Correct iOS convertToUI

eraserhd:
-convertToGL: and -convertToUI: were not symmetric, there were two issues:
1. convertToUI: worked in points, but then attempted to convert the location (already in points) to points from pixels.
2. convertToUI: for Landscape Right mode did something both inconsistent with convertToGL and wrong.


eraserhd:
Pull request closed, will resubmit on develop.

PR #30, opened 2011-01-28
[CocosDenshion] restrict the range of effectsVolume & backgroundMusicVolume

walzer:
If i set the property
[[SimpleAudioEngine sharedEngine].effectsVolume = -0.1f
or
[[SimpleAudioEngine sharedEngine].effectsVolume = 100.0f

then
[[SimpleAudioEngine sharedEngine] playEffect: @"dp1.caf"];

It output noise, no matter the effect is in caf format or wav format.

I think it's caused by the range of effectsVolume 0.0~1.0f. So I modify SimpleAudioEngine.m to make it safer.

### DIFF
--- a/CocosDenshion/CocosDenshion/SimpleAudioEngine.m
+++ b/CocosDenshion/CocosDenshion/SimpleAudioEngine.m
@@ -194,6 +194,9 @@ static CDBufferManager *bufferManager = nil;
 
 -(void) setBackgroundMusicVolume:(float) volume
 {
+	volume = volume < 0 ? 0.0f : volume;
+    volume = volume > 1 ? 1.0f : volume;
+	
 	am.backgroundMusic.volume = volume;
 }	
 
@@ -205,6 +208,9 @@ static CDBufferManager *bufferManager = nil;
 
 -(void) setEffectsVolume:(float) volume
 {
+	volume = volume < 0 ? 0.0f : volume;
+    volume = volume > 1 ? 1.0f : volume;
+	
 	am.soundEngine.masterGain = volume;
 }	
 


ghost:
What device and iOS combination does this occur on?  It is actually desirable to have effects volume go above 1.0 as this can allow the user to amplify the sounds so they are heard over their music.  Setting effects volume to 2.0 is supposed to give a +6dB boost so setting it to 100 is just crazy.  However, it seems that in 4.x Apple clamps the value to 1.0 any way.  The OpenAL spec is not clear on what the max. value should be only what the amplitude gain is for doubling the value.

Also this should be fixed at the CDSoundEngine level.  Making the change in SimpleAudioEngine doesn't fix things for people who only use CDAudioManager or CDSoundEngine.

walzer:
The bug occurs on iOS 4.2 simulator. I test the volume at 2.0, it works correctly. But 100 is too large.
The defect apears when someone write such code:
-(void) subVolume
{
    float volume = [[SimpleAudioEngine sharedEngine] effectsVolume];
    [SimpleAudioEngine sharedEngine].effectsVolume = volume - 0.1f; 
        // this may decrease to negative value
}

You're right. It must be fixed at the CDSoundEngine level.

ghost:
I'll fix it in the next batch of changes I do.  However, as I said above - I think on the device Apple are clamping the value to 0-1 anyway.

walzer:
well, waiting for your fix :)

ricardoquesada:
@steve: Should I merge it in the cocos2d repository ?

ghost:
@riq - no I'll fix it properly with my next batch of changes.  It is a simulator only issue that hasn't bothered anyone for 18+ months so I don't think there is any urgency.

ricardoquesada:
@steve: Ok, thanks. I'll wait for your patch then.

PR #32, opened 2011-02-02
Suppress warnings in CCTexturePVR

splhack:
Hi, I fixed for clang warnings with between '%lu' and NSUInteger arguments.

warning: conversion specifies type 'unsigned long' but the argument has type 'NSUInteger' (aka 'unsigned int') [-Wformat]


splhack:
It was fixed by this commit. Thank you!
https://github.com/cocos2d/cocos2d-iphone/commit/403fe2ff79d489deca6ff8d3f098e7e02dec3afd

PR #34, opened 2011-02-14
Coordinate conversion fixes

eraserhd:
This corrects coordinate conversions so that convertToGL and convertToUI are symmetric and also fixes handling of one landscape orientation, making it also symmetric (as well as correct).

### DIFF
--- a/cocos2d/Platforms/iOS/CCDirectorIOS.m
+++ b/cocos2d/Platforms/iOS/CCDirectorIOS.m
@@ -526,8 +526,7 @@ CGFloat	__ccContentScaleFactor = 1;
 			uiPoint = ccp(glPoint.y, glPoint.x);
 			break;
 		case CCDeviceOrientationLandscapeRight:
-			// Can't use oppositeX/Y because x/y are flipped
-			uiPoint = ccp(winSize.width-glPoint.y, winSize.height-glPoint.x);
+			uiPoint = ccp(oppositeY, oppositeX);
 			break;
 	}
 	return uiPoint;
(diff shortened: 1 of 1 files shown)


ricardoquesada:
@eraserhd thanks for the patch. 
What's the best way to test your fix ? Do you have an small test case ?
Currently I don't have any test case so I'm not sure what is broken right now. Thanks.

eraserhd:
Based on the issues we saw, we were pretty sure that people could only be converting coordinates in one direction - to GL from UI.  We needed to do the opposite direction for iCuke to produce the accessibility XML that iCuke needs to interact with the application.

Patch 9846 fixes issues that we had when trying to use iCuke with Retina displays.  convertToUI converts _points_ in GL space to _points_ in UI space, so the scaling factor shouldn't be applied unless someone later wants to convert the coordinate to pixels.

Patch 30b0 happened because we noticed that convertToUI and convertToGL had asymmetrical logic for landscape right.  We didn't actually need this at the time, (using landscape left in simulator), but we reasoned that for any CGPoint p, [director convertToUI:[director convertToGL:p]] should equal p.  The logic which was there (along with the comment) produces strange out-of-bounds coordinates.

To test, I would just pick some random points and verify that [director convertToUI:[director convertToGL:p]] == p in each orientation.

ricardoquesada:
Thanks. I'll add the test you mentioned.

ricardoquesada:
thanks. Added DirectorTest that tests the above mentioned scenario.
- RetinaDisplay commit merged
- "device-right-oriented" commit rejected. It seems that it was working OK and the 2nd commit was breaking it.

PR #35, opened 2011-02-15
Loop sound effects

Omniwheel:
This adds the ability to loop sound effects when using SimpleAudioEngine.


ghost:
Your code has a bug.  Try setting up a loop and then play 32 effects.  The loop will be terminated.  If you want to do looping you have to use a CDSoundSource as this locks the underlying source and prevents it getting reused.

PR #38, opened 2011-03-02
Updated FontLabel class

chiefy:
We ran into a problem with FontLabel where text was not properly wrapping (it would wrap on the apostrophe character, and sometimes hyphen etc.) I saw someone had forked and fixed the issue, and he informed me that he is the original author (kballard), and now maintaining it under a different fork.


ricardoquesada:
Thanks. Merged.
https://github.com/cocos2d/cocos2d-iphone/commit/1d19393664fe92b6786917dba16547b39a2e10f1

PR #41, opened 2011-03-17
Sprite cache naming

mrcharles:
This modification enables the use of non-unique sprite names across multiple spritesheets without name collisions in the sprite frame cache. It's a little end-user heavy, since they have to remember and keep track of whatever tag they give to a particular set of sprite frames. But it's adequate and it solved a problem I was having (I'm using a lot of processed sprites, and this modification was easier than making sure every sprite was guaranteed a unique name). 

I considered implicitly adding the name of the spritesheet/png/plist to the cache key, but decided against it in order to facilitate proper backwards compatibility. 

All tests in SpriteTest run as they did in the master branch, however, the alias test doesn't function in the master branch so I was unable to confirm compatibility, though I'm pretty sure I have it covered. 

Other change is that I added a warning to the logs when a sprite frame cache entry was clobbered. This log mimics the one set up for testing against duplicate aliases. 

Overall this change is probably not as robust as it probably should be for a release, but I think it's a solid foundation to build upon.

### DIFF
--- a/cocos2d/CCSpriteFrameCache.h
+++ b/cocos2d/CCSpriteFrameCache.h
@@ -55,10 +55,15 @@
  */
 +(void)purgeSharedSpriteFrameCache;
 
+/** Generates the key that will be used to index in to the cache dictionary
+ can be used to build the key from outside frame cache
+ */
++(NSString*) makeCacheKeyFromName: (NSString*)name basekey:(NSString*)key;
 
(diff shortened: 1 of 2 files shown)


araker:
I agree that eventually there should be a system that accepts unique names per spritesheet and not per all spritesheets. Like you say, your solution with tags adds a bit of stress to the programmer, because the tags need to be remembered. I find it not user-friendly enough to be included into cocos.

PR #42, opened 2011-03-17
CCSpriteFrame shouldn't retain texture if it's possible.

psineur:
Right now in cocos2d CCSpriteFrame retains texture with which it is created.
This leads to some kind of leaking texture - it will not be dealloced until CCSpriteFrame's retaining it will not be purged from CCSpriteFrameCache.

But i think it's possible and usefull to keep CCSpriteFrame's in CCSpriteFrameCache without retaining the texture.
My solution isn't absolute lazy-texture-alloc, but it's tiny and helps me to avoid memory warning crashes without purging CCSpriteFrameCache.

Aso i should mention, as TBBle said:
"That'd make the wiki's description of purging unused textures actually correct."

Original topic was here: http://www.cocos2d-iphone.org/forum/topic/13732


psineur:
Story behind this patch:

I was just trying to fix memory crashes in next update of iTraceur, i started adding many
`[[CCSpriteFrameCache sharedSpriteFrameCache] removeSpriteFramesFromFile: @"someFile.plist"];`
calls in my Menus Node's deallocs, but this fix was faster and worked fine for me.
I've tested it in iTraceur 1.3 on 4.2.1 iPod Touch 2Gen - no more crashes.

Also i tested it in latest cocos2d develop branch state with SpriteTest and ZwoptexTest on iPhone Simulator 4.2
- Everything worked the same as before the patch.

psineur:
Ooops, looks like i need to close this pull request first. 
Will reopen later if needed - tell me if you don't want this change(CCSpriteFrame not retaining texture) in Cocos2D.

PR #43, opened 2011-03-18
Fix for issue 1137 (archive support with xcode 4)

josh-h:
The simple change required to support static library archive support with Xcode4.


ricardoquesada:
merged. thanks.

PR #46, opened 2011-04-01
fix for unscheduling updates

thoechsmann:
Hi Riq,

my last pull request was for the wrong branch. Here is the correct one.

It fixes this bug: http://code.google.com/p/cocos2d-iphone/issues/detail?id=1144.

Thomas


thoechsmann:
I found a problem with my recent implementation. I will send another pull request and close this one.

PR #47, opened 2011-04-01
fixed unscheduling of updates

thoechsmann:
Hi Riq,

this should fix problems with unscheduling updates.

If we are in the tick method of the CCScheduler we do not delete updates from the hash immediately, but mark them for deletion. If they are marked, they will no longer be called in the current tick and they will also be deleted at the end of the tick.

If we are not in the tick function, updates are deleted immediately.

Thomas

### DIFF
--- a/cocos2d/CCScheduler.m
+++ b/cocos2d/CCScheduler.m
@@ -43,8 +43,8 @@ typedef struct _listEntry
 	TICK_IMP impMethod;
 	id		target;				// not retained (retained by hashUpdateEntry)
 	int		priority;
-	BOOL	paused;
-	
+	BOOL	paused;	
+    BOOL    markedForDeletion; // selector will no longer be called and entry will be removed at end of the next tick
 } tListEntry;
 
 typedef struct _hashUpdateEntry
@@ -198,6 +198,7 @@ static CCScheduler *sharedScheduler;
 		currentTarget = nil;
 		currentTargetSalvaged = NO;
 		hashForSelectors = nil;
+        updateHashLocked = NO;
 	}
 
 	return self;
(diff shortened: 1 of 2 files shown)


cocos2d/CCScheduler.m
```diff
@@ -401,11 +403,18 @@ static CCScheduler *sharedScheduler;
...
-#if COCOS2D_DEBUG >= 1
 	tHashUpdateEntry * hashElement = NULL;
 	HASH_FIND_INT(hashForUpdates, &target, hashElement);
-	NSAssert( hashElement == NULL, @"CCScheduler: You can't re-schedule an 'update' selector'. Unschedule it first");
+    if(hashElement)
+    {
+#if COCOS2D_DEBUG >= 1        
+        NSAssert( hashElement->entry->markedForDeletion, @"CCScheduler: You can't re-schedule an 'update' selector'. Unschedule it first");
 #endif	
+        // TODO : check if priority has changed!
```
psineur, line 413: Is this TODO still necessary?

psineur:
How does it work right now, without this patch? It crashes if update: selector is uncheduled from itself?
Looks like it's only for scheduleUpdate and not for scheduleSelector, right?

thoechsmann:
The TODO is still required. With the patch it is not possible to unschedule an update and reschedule it with a different priority during the same tick (when it was not yet removed from the hash).

Currently (without the patch) it can crash if you unschedule un update from the update of another object.

thoechsmann:
The TODO is still required. With the patch it is not possible to unschedule an update and reschedule it with a different priority during the same tick (when it was not yet removed from the hash).

Currently (without the patch) it can crash if you unschedule un update from the update of another object.

thoechsmann:
The TODO is still required. With the patch it is not possible to unschedule an update and reschedule it with a different priority during the same tick (when it was not yet removed from the hash).

Currently (without the patch) it can crash if you unschedule un update from the update of another object.

ricardoquesada:
merged for v1.0
https://github.com/cocos2d/cocos2d-iphone/commit/4ee93f7cb77ee08a94e2e8918c150c12d63b0bf5

PR #50, opened 2011-04-14
Added isRetina method for CCTexture2D. Code from Patrick Wolowicz

psineur:
Tiny change that will allow CCTexture2D to know, was it loaded for retina mode (from foo-hd.png) or not (from foo.png) when user asked it to load from foo.png

Can be usefull to decrease resources, if universal resources can be used for iPad/iPhone/iPhone retina.
Example: https://github.com/psineur/CCMenuAdvanced/commit/5a6815e9c80c3fc30d1877d643ce239386858c28

This is about issue #1118 , but without automatic upscaling, just give user a chance to do it manually, by knowing is texture loaded from hd resource or not.
Original Patrick's code taken from here: http://www.cocos2d-iphone.org/forum/topic/13935

### DIFF
--- a/cocos2d/CCTexture2D.h
+++ b/cocos2d/CCTexture2D.h
@@ -121,6 +121,7 @@ typedef enum {
 	GLfloat						maxS_,
 								maxT_;
 	BOOL						hasPremultipliedAlpha_;
+	BOOL						isRetina;
 }
 /** Intializes with a texture2d with data */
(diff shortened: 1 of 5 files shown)


ricardoquesada:
to included in v1.1

ricardoquesada:
Thanks. Fixed in develop branch.
The patch was not merged, but was a similar code was added.

PR #52, opened 2011-04-21
CCParticleSystemAnimation class -- animated particle quads

alanf:
This is the first pull request I've created for cocos2d, so I would appreciate any feedback you had.

Not sure what to do with the version numbers, and if I'm following all the proper conventions.

This functionality is for a feature I'm adding to a Pocket Gems game; it's to provide even more flexibility when creating Particle Systems.


alanf:
I just read the FAQ, so I'll close this and resubmit per the instructions there

ricardoquesada:
Alan, It is OK to send me pull request.
What does the FAQ say ? Does it say something regarding not sending "pull requests" ? If os, I should fix it

alanf:
Thanks Ricardo. I branched from master instead of the dev branch. I'll correct that and resubmit the request.

leegrey:
Hi. I'm interested in using your animated particle class. I was wondering if you had an example of how to use it. I havn't had any luck so far, just calling setAnimation. Should I expect this to work with the main release branch of Cocos2D? Any idea when it might appear in the main branch?

leegrey:
Ah. My emissionRate is set to zero as I am adding bursts of particles manually. This meant that the conditional in the update function was never passing.

PR #53, opened 2011-04-22
Particle animation

alanf:
Extends CCParticleSystemQuad to draw quads which are animated (if a CCAnimation is set)


araker:
Closing this one, v1.1 already has a animated particle system.

PR #54, opened 2011-05-02
Skew

chapados:
We added skew support to cocos2d for our last game.  This patch includes support for skew transforms, actions, and action tests.  We extracted these changes from our public 'booyah' branch and added the changes on top of the latest master, to remove extra stuff that is not necessarily relevant to cocos in general.  This goes against the standard convention of re-writing public commits, but no one has forked our branch, so this should be fine.  We think this is a useful addition to cocos2d and hope you'll consider including this feature.


ricardoquesada:
Thanks. Very useful additions. Merged
https://github.com/cocos2d/cocos2d-iphone/commit/fa02b8ac43d48919127e6bca7398f3efa96ed95b

PR #57, opened 2011-05-18
CCTextureCache releases the returned texture in 'addImage' rather than autoreleasing.

glimajr:
CCTextureCache's 'addImage' method currently releases the generated texture before returning it.

In a single-threaded application, this is no problem as the TextureCache itself keeps the retain count at 1. However, in a multithreaded environment, the TextureCache may be freed before the caller of addImage (or addImageAsync) receives the returned texture.

This fork simply autoreleases the texture prior to returning it.

### DIFF
--- a/cocos2d/CCTextureCache.m
+++ b/cocos2d/CCTextureCache.m
@@ -280,7 +280,7 @@ static CCTextureCache *sharedTextureCache;
 			else
 				CCLOG(@"cocos2d: Couldn't add image:%@ in CCTextureCache", path);
 			
-			[tex release];
+			[tex autorelease];
 		}
 		
 		else {
@@ -297,7 +297,7 @@ static CCTextureCache *sharedTextureCache;
 			else
 				CCLOG(@"cocos2d: Couldn't add image:%@ in CCTextureCache", path);
 			
-			[tex release];			
+			[tex autorelease];			
 		}
 
 		// Only in Mac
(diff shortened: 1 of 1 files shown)


ricardoquesada:
Could you describe in more detail the scenario that this patch tries to solve ? 
Are you releasing the texture from another thread ? which method are you calling ? Are you using cocos2d-mac ?
Thanks!

glimajr:
It's basically really hard to reproduce the bug since it's concurrency-based. But the basic idea is:
1. I call 'addImageAsync'.
2. This turn calls 'addImageWithAsyncObject' in a background thread.
3. 'addImageWithAsyncObject' then calls 'addImage', still in the background thread.
4. 'addImage' will alloc a new texture, add it to the TextureCache, release it, and then return it to 'addImageWithAsyncObject' (which is running in the background thread).
5. Something in the main thread calls any of the methods that will clear the cache of textures, such as 'removeUnusedTextures' or 'purgeTextureCache'. This causes the texture to be dealloc'ed because it isn't being retained by an autorelease pool (the only retain is by the TextureCache).
6. In the background thread, 'addImageWithAsyncObject' then calls 'performSelectorInMainThread' with the pointer to the now-freed texture.
7. Back in the main thread, the callback I supplied to 'addImageAsync' is called but with a pointer to freed memory. I get an EXC_BAD_ACCESS when I try to access it.

glimajr:
The easy way to simulate something like this, is to add line 158 in CCTextureCache.

```
156         // load / create the texture
157         CCTexture2D *tex = [self addImage:async.data];
158         [[CCTextureCache sharedTextureCache] removeAllTextures]
159         // The callback will be executed on the main thread
160         [async.target performSelectorOnMainThread:async.selector withObject:tex waitUntilDone:NO];
```

ricardoquesada:
thanks. merged in "develop" branched.

PR #58, opened 2011-05-18
CCNode pointAtPosition

derekvanvliet:
I added a method to CCNode that can be used to orient it such that the positive x axis is pointing towards a relative position.

Also included is a demo in cocosnodeTest.


araker:
Thanks for the patch. It's a handy little method, but the functionality is too small to be included into CCNode. Also ccpAngle(a,b) does the same thing (though it's two lines of code instead of one).

PR #62, opened 2011-05-25
Implement CCRGBAProtocol for CCProgressTimer

alco:
Before this change, it was impossible to apply an action
like CCFadeIn or CCFadeOut to the 'sprite' property of
CCProgressTimer, because CCProgressTimer overrides the 'draw'
method.

With this change applied, we can now write

```
CCProgressTimer *ptimer = ...;
ptimer.opacity = 0;
[ptimer runAction:[CCFadeIn actionWithDuration:1.0]];
```

and it's going to work.

Signed-off-by: Alexei Sholik alcosholik@gmail.com

### DIFF
--- a/cocos2d/CCProgressTimer.h
+++ b/cocos2d/CCProgressTimer.h
@@ -50,7 +50,7 @@ typedef enum {
  The progress can be Radial, Horizontal or vertical.
  @since v0.99.1
  */
-@interface CCProgressTimer : CCNode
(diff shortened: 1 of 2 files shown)


manuqiao:
Any detail?Which files to be modified?

araker:
Thanks, fixed..

https://github.com/cocos2d/cocos2d-iphone/commit/8866f46c4169169dce89997463a75f58a8bcb305

PR #63, opened 2011-05-26
assert on ccArrayDoubleCapacity

rolandoam:
it checks that the realloc did actually worked (otherwise, arr->arr is set to a NULL pointer)


ricardoquesada:
merged. thanks.

PR #68, opened 2011-06-16
added the facility to overide resource loading to a custom NSBundle

nitaigao:
Hi guys,

We had a requirement to be able to load resources from a custom location outside of the main application bundle on a mac, this is good for level editors and the like.

I've added a method that sets the bundle that CCFIleUtils pulls from. By default this is of course [NSBundle mainBundle] but this can now be overridden using:

[CCFileUtils setBundle:_bundle_];

All the best and thanks for doing such a good job on the project.

Nick

### DIFF
--- a/cocos2d/Support/CCFileUtils.m
+++ b/cocos2d/Support/CCFileUtils.m
@@ -32,6 +32,7 @@
 #import "../ccConfig.h"
 
 static NSFileManager *__localFileManager=nil;
+static NSBundle *__bundle=nil;
 
 // 
 NSInteger ccLoadFileIntoMemory(const char *filename, unsigned char **out) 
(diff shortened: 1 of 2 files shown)


nitaigao:
Cool, i've made those changes.

cocos2d/Support/CCFileUtils.m
```diff
@@ -94,8 +95,19 @@ NSString *ccRemoveHDSuffixFromFile( NSString *path )
 
 +(void) initialize
 {
-	if( self == [CCFileUtils class] )
+	if( self == [CCFileUtils class] ) {
 		__localFileManager = [[NSFileManager alloc] init];
+    __bundle = [NSBundle mainBundle];
```
haqu: this should be retained too: `__bundle = [[NSBundle mainBundle] retain];`

cocos2d/Support/CCFileUtils.m
```diff
@@ -94,8 +95,19 @@ NSString *ccRemoveHDSuffixFromFile( NSString *path )
...
+  }
+}
+
+- (void)dealloc {
+  [__bundle release];
+  [super dealloc];
+}
+
++ (void)setBundle:(NSBundle*)bundle {
+  __bundle = [bundle retain];
```
haqu, line 113: in case bundle is already set, we should release it first:

```
if (__bundle) {
  [__bundle release];
}
__bundle = [bundle retain];
```

nitaigao:
Thats a very good point, thats now done.

PR #70, opened 2011-06-20
Change converting touch location view to OpenGL view.

emeitch:
If you create a Cocos2d applications in the complex UIView hierarchy, UITouch's "view" method returns a GLView not always. 
Next line, locations are converted for GLView.
So, please specify explicit the GLView.


rsanchezsaez:
I think this issue is very important if you plan to use cocos2d together with UIKit elements. This fix is also included in my pull request.

ricardoquesada:
What is the easies way to test it ?

araker:
Thanks, implemented.

PR #83, opened 2011-07-20
Feature/2x

kstenerud:
Makes cocos2d search for @2x or -hd versions of an image when in retina mode.


ricardoquesada:
The idea is support only one extension to make the code simpler and faster... actually it should be possible to support @2x extension by adding the following macro in the "predefined macros" in Xcode:
CC_RETINA_DISPLAY_FILENAME_SUFFIX=@"@2x"

PR #84, opened 2011-07-20
Feature/update duration

kstenerud:
Gives actions the opportunity to change their durations just prior to running.

This gives much greater opportunity for action re-use, especially in complex action structures where one part of an action tree changes between action invocations. In this case, it's much more efficient to just re-use the action tree with recalculated duration rather than rebuilding the whole thing.

Once this base is in place, I'll submit updateDuration aware versions of sequence and spawn.


ricardoquesada:
thanks. scheduled for v1.1 / 2.0

PR #85, opened 2011-07-20
Feature/pause all actions

kstenerud:
Allows all actions to be paused, returning a list of targets that were paused.

An example usage would be:
- Pause all actions
- Surface a UI node using actions (zoom in, slide in, whatever)
- When the UI node has been dismissed, resume all actions that were paused.


ricardoquesada:
thanks. scheduled for v1.1 / 2.0

araker:
Implemented in 1.1 as well.. 

https://github.com/cocos2d/cocos2d-iphone/commit/36b1ec1414d82a39e3f128318842952bab976599

PR #86, opened 2011-07-20
Feature/pause and unschedule

kstenerud:
A clean way to remove all selectors that are not required by the system.

This makes leaving a scene with scheduled selectors easier since you can just stop all selectors with non-system priority and not have to worry about what node scheduled what, and what's still active.

I use it in my pause screen for exiting the current game, for which the pause screen has no knowledge of what's still running.


ricardoquesada:
thanks. I'll study the patch a bit further.... I'm not sure if it might break something.

PR #87, opened 2011-07-20
Add support for named tags

kstenerud:
Named tags are a prerequisite to adding functional testing support via KIF https://github.com/square/KIF


ricardoquesada:
Thanks. I'm not sure I would like to add it to the main branch... perhaps it could be added as an extension.

kstenerud:
Unfortunately, it requires an ivar, so it can't be added as a category, otherwise I would have done that ;-)

ricardoquesada:
I think it is possible to add ivars (or properties) as extensions using llvm 2.0 (or newer). But I might be wrong

PR #90, opened 2011-07-25
ARC-compatible header files

jerrodputman:
A few changes that make Cocos2D's header files ready to be included in ARC projects. The entire procedure and reasoning behind it can be found here: http://www.tinytimgames.com/2011/07/22/cocos2d-and-arc/

Basically, while the Cocos2D source is not ARC-compatible, by including the framework as a static library or cross-project reference (or even manually disabling ARC on each source file in the framework), Cocos2D can still be used in an ARC-compiled app. A few minor changes to the header files were necessary to make this work, since they will be included in ARC-compiled source files.

The changes should also be backwards-compatible with compilers that don't recognize ARC.


ricardoquesada:
Thanks. ARC support will be added in cocos2d 2.x

cocos2d:
Thanks. Merged both in v1.x and 2.x.

PR #93, opened 2011-08-25
Scaling of low-res images on Retina devices

nanotech:
This patch attempts to solve [#1118](http://code.google.com/p/cocos2d-iphone/issues/detail?id=1118) in a reasonably clean way. The main idea is to separate the concept of screen pixels from texture data pixels. By not assuming they always have a 1:1 ratio to each other (and thus a 2:1 ratio to retina points), low-res images can be given a different ratio to keep them at a constant physical size.

I've done this with a set of macros that bypass point-pixel unit conversion on low-res textures, giving their pixels a 1:1 ratio to points.

Sprites, sprite sheets, and bitmap fonts have been modified to use the new macros and seem to be working for what I'm using them for, but I've almost certainly missed some things. I only took a cursory look at TMX maps and haven't implemented anything for them yet.

Thanks for the great framework!

### DIFF
--- a/cocos2d/CCLabelBMFont.m
+++ b/cocos2d/CCLabelBMFont.m
@@ -567,8 +567,8 @@ typedef struct _KerningHashElement
 		}
 		
 		float yOffset = configuration_->commonHeight_ - fontDef.yOffset;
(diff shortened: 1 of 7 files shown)


drewish:
Found this when I was looking at scaling of TMX maps: http://wasabibit.com/WasabiBit/Dev_Notes.html

slembcke:
This is a really old pull request, but you might be happy to know that this is fixed in 3.0. Textures all have their own independent content scale, and the screen's content scale can be setup however you want as well. (ex: for pixel doubled games with pixel art.)

PR #95, opened 2011-09-04
Adding in-memory TMX map creation

slycrel:
Added convenience methods to take an XML string and create an in-memory
TMX map from it rather than reading it from a file.  Also added a test for this.  See issue #1239


slycrel:
Hey Araker, any chance of this seeing a mainline merge anytime soon?  And if so, did it miss the deadline for the 1.1 beta?

araker:
It's only since yesterday that I have permission to merge pull requests into cocos main repository.  So I haven't looked at the various pull requests till now. I'll look into your patch in the next couple of days.

slycrel:
No worries.  Sounds like you are newer at this than I thought, thanks for the quick response!

PR #109, opened 2011-10-03
Issue 1254, freeze in layoutSubviews call 

araker:
This fix ensures that OpenGL frame buffers, render buffers are destroyed and recreated when needed. Before the buffers were merely resized. This can lead to a lengthy freeze when starting up.

Tested with bug 914 case.


araker:
Please  wait with integrating this request, it seems it can generate a 0x502 error in OpenGL. I'll do some more testing.

PR #122, opened 2011-11-01
Allow CCNode to notify its parent when it changes size or scale

kstenerud:
In order to make container nodes able to intelligently respond to changes in its children, there needs to be a callback mechanism for the child to notify the parent of a change.

An example would be a grid which resizes itself to accomodate changes in its children and maintain spacing.


psineur:
Why just not use KVO? It does the same thing - but you're doing it outside, so it needs no modifications in Cocos2D.

PR #124, opened 2011-11-11
Modify CCLabelBMFont for multiline alignment

pingpongboss:
Left, Center, and Right alignments.
Auto word wrap based on given width.

Based on https://github.com/pingpongboss/CCLabelBMFontMultiline

### DIFF
--- a/cocos2d/CCLabelBMFont.h
+++ b/cocos2d/CCLabelBMFont.h
@@ -137,6 +137,13 @@ enum {
 {
 	// string to render
 	NSString		*string_;
+    
+    // initial string without line breaks
+    NSString *initialString_;
+    // max width until a line break is added
+    float width_;
+    // alignment of all lines
+    UITextAlignment alignment_;
 	
 	CCBMFontConfiguration	*configuration_;
 
@@ -152,6 +159,10 @@ enum {
  */
 +(void) purgeCachedData;
 
+@property (nonatomic,copy,readonly) NSString *initialString;
+@property (nonatomic,assign,readonly) float width;
+@property (nonatomic,assign,readonly) UITextAlignment alignment;
+
 /** conforms to CCRGBAProtocol protocol */
 @property (nonatomic,readwrite) GLubyte opacity;
 /** conforms to CCRGBAProtocol protocol */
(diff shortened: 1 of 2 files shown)


ricardoquesada:
Hi PingPong,

Many thanks for the patch. I have just tested, and I like the new functionality. 
But it has one major problem with the performance. Every time the label is updated, all the characters all deleted and recreated. This is too expensive for labels that might be updated every frame.

Do you think it would be possible to fix that issue ? Thanks again.

pingpongboss:
Hi Ricardo,

Ah, I haven't thought about the performance ramifications. The reason why I remove all the characters is that if I don't, some funky stuff happens with whitespaces and I either can't auto-line-break correctly (step 1), or I can't count the number of characters correctly for alignment (step 2). It has something to do with createFontChars reusing the old children in some bad way.

I will give it a try to make it work, but if it's possible maybe someone could take a look at createFontChars to see what's going wrong.

ricardoquesada:
Hi PingPong,

I developed "createFontChars" so I can help you with that. Could you describe a bit further the issue and how to reproduce it ? 
Thanks.

pingpongboss:
Sure, let me play around with disabling removeAllChildren and observe the side-effects again.

ricardoquesada:
Ok. Thanks.

On 11 November 2011 21:26, Mark Wei
reply@reply.github.com
wrote:

pingpongboss:
Hey Ricardo,

Sow I thought it was impossible for me but I think I've done it. Can you check out the latest changes and see if they work for you.

The main problem I was having was that I was under the assumption that doing a `for(CCSprite *child in children_)` would go through the children in order. But instead, I had to go by the tag to go in order. I also had to take care of the cases where a child would be missing for a specific tag, for example: `[children_ count]` returns 3, but only children with tags 1, 2, and 4 exist. Lastly I had to skip over children who were not `visible`.

I'm not sure where these side-effects came from, but I'd guess it comes from `createFontChars`'s optimizations when you give the label a new, similar string to render.

Thanks,

pingpongboss:
ah, btw you can call me Mark

ricardoquesada:
Thanks Mark,

I will test it today.

pingpongboss:
Cool, let me know if there are any problems

ricardoquesada:
Thanks Mark.
Patch merged in gles20 branch.
https://github.com/cocos2d/cocos2d-iphone/commits/gles20

PR #125, opened 2011-11-20
CCAction/ActionManager bugfix

jaeschliman:
here's an initial bugfix with a test for issue 945 as spoken about here:
http://www.cocos2d-iphone.org/forum/topic/22955
There's not much to it, but I'd love to hear feedback before I clean it up/make sure all actions work with it properly etc.
There's a new target SequenceTest in the mac project that demonstrates adding and stopping several
sequence/repeat actions independently on a single node.
Lemme know what you think!
-j

### DIFF
--- a/ActionTest copy-Info.plist
+++ b/ActionTest copy-Info.plist
@@ -1,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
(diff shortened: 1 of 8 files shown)


araker:
I've implemented your test case for the repeatForever and fixed Repeatforever and the action manager in a similar way as you proposed. You're also mentioned in the authors, for the test case and the bugfix

https://github.com/cocos2d/cocos2d-iphone/commit/16f4844b54beaa73d5a2d6cfc4e4208f6b591bcf 
https://github.com/cocos2d/cocos2d-iphone/commit/d11d1ed9b54aeb1ce2782262d24aa1fc52a3945c

What exactly is the problem with ActionEase?

PR #126, opened 2011-11-29
CCEaseInOut fix

akotiranta:
Fixed the CCEaseInOut to be able to handle arbitrary rate values above 1. Previous implementation worked only for integer values. The test cases are changed to describe the problem, previous implementation would fail on the cases now being in the test.


ricardoquesada:
Hi, thanks. 

This pull request doesn't have any patch attached. Could you re-send a new pull request or attach the patches? thanks.

akotiranta:
Sorry, I hope my new pull request is better.

PR #128, opened 2011-11-29
In CCEventDispatcher delegate is not released when freed

akotiranta:
OSX version of CCEventDispatcher has a bug when releasing all delegates the delegates are not freed like when they are released individually. Fortunately this piece of code is rarely if ever used, but it's good to be fixed anyway.


ricardoquesada:
Thanks. fixed on v2.0-beta2

PR #129, opened 2011-12-13
Backported CCLabelBMFont from gles20 branch.

akotiranta:
This backport is tested on iPhone 3G and an iPod with retina display using LabelTest from the cocos2d project. Also it works on my current game so hopefully there are no issues. I changed as little as possible.

I also added arm6 to the cocos2d, FontLabel and LabelTest targets, I think arm6 should be added to all targets for 1.x cocos2d versions since only reason to use 1.x cocos2d is to support older handsets.


ricardoquesada:
thanks

araker:
Merged it without the target changes. Arm6 is already available in 1.1 for all targets.

https://github.com/cocos2d/cocos2d-iphone/commit/98ab59e176a84f548216079bda5aec625e94833d

PR #132, opened 2011-12-19
Hierarchical per-node time scaling

lhunath:
This allows you to specify a time scale on a node in the scene, which causes all actions and schedules on it and all its children to get scaled in time.  This is vital to slow down certain parts of a scene; eg. all the game but not the menus.


lhunath:
You probably don't want to apply 711da56

ricardoquesada:
Thanks.
Interesting approach. Have you done any performance tests? How much does it affect the runtime.

For cocos2d v2.0 it is also possible to alter the speed of a tree of nodes by using multiple schedulers and/or action managers. I will create an example soon.

lhunath:
I haven't; but I don't expect the cost to be high, especially when the developer isn't using it explicitly.  Then it's just the cost of respondsToSelector:.  Worth testing, but I can't make any time for it lately, I'm afraid.

araker:
This is a interesting approach indeed, but I rather keep the action manager as agnostic / simple as possible. 

A alternative is to use the new CCTargetedAction, which gives a target to a action to make it possible that a node can run actions that aren't applied to the node itself. Using a sequence of targeted actions wrapped in a CCSpeed would do the same as this extension..

lhunath:
I fail to see how a CCTargetedAction could affect all actions and schedules previously applied to a tree of nodes.

lhunath:
Given that action managers manage time for a target, I think it's acceptable that it also manages the scaling of that time allocation.  I wouldn't call that bloat.

I haven't tried the multiple schedulers/action managers approach yet, I don't know what it would involve, but ideally, it shouldn't require all nodes in a tree or all actions and schedules applied to them to perform some special logic such as wrapping actions or assigning special managers to each node in the tree.

araker:
Using CCTargetedAction would mean flatten the tree, wrap all actions in a CCTargetedAction and put them in a sequence/spawn which is wrapped in a CCSpeed. Changing the rate of the speed action would then give the same result. I have to admit, it's certainly not as convenient as your code.  

Using a seperate manager for a certain hierarchy of nodes would be a interesting approach as well.

PR #145, opened 2012-01-26
A fix for supporting CCSprite with children added to a CCSpriteBatchNode

jpsarda:
Only the commit to fix the CCSprite.m is important. I'm sorry I'm new to GitHub and I was not able to send a pull request with only this commit. I think I should have created a new branch maybe but I'm not sure to understand branches yet.

More infos on the forum : http://www.cocos2d-iphone.org/forum/topic/28351


ricardoquesada:
Thanks.
I fixed it in v2.0 using another approach:
https://github.com/cocos2d/cocos2d-iphone/commit/cb5df64a8366c414a9f873f268f74932c459f304

Leaving pull request open since it is not fixed in v1.1 yet

araker:
I've integrated this patch in 1.1, but didn't  use the pull request. 

https://github.com/cocos2d/cocos2d-iphone/pull/150

@jpsarda, you're mentioned as the author of the patch. I'm closing this patch now.

jpsarda:
Thx guys.

PR #155, opened 2012-02-10
Gles20

Panajev:
Added uniform caching mechanism. Jumpy Lens3D's redundant calls down to 15 (30 s period).

### DIFF
--- a/cocos2d/CCAtlasNode.m
+++ b/cocos2d/CCAtlasNode.m
@@ -124,7 +124,10 @@
 
 	ccGLBlendFunc( blendFunc_.src, blendFunc_.dst );
 
-	glUniform4f( uniformColor_, color_.r / 255.0f, color_.g / 255.0f, color_.b / 255.0f, opacity_ / 255.0f );
+	//glUniform4f( uniformColor_, color_.r / 255.0f, color_.g / 255.0f, color_.b / 255.0f, opacity_ / 255.0f );
+    
+    ccColor4F col = {color_.r / 255.0f, color_.g / 255.0f, color_.b / 255.0f, opacity_ / 255.0f };
+    [shaderProgram_ loadUniform:@"u_color" withValue:&col withType:@"vec4f"];
 
 	[textureAtlas_ drawNumberOfQuads:quadsToDraw_ fromIndex:0];
 }
--- a/cocos2d/CCDrawingPrimitives.m
+++ b/cocos2d/CCDrawingPrimitives.m
@@ -75,7 +75,7 @@ void ccDrawPoint( CGPoint point )
 	ccVertex2F p = (ccVertex2F) {point.x, point.y};
 
 	ccGLEnableVertexAttribs( kCCVertexAttribFlag_Position );
-	ccGLUseProgram( shader_->program_ );
+	ccGLUseProgram( shader_ );
 	ccGLUniformModelViewProjectionMatrix( shader_ );
 
 	glUniform4f( colorLocation_, color_.r, color_.g, color_.b, color_.a );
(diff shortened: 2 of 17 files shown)


ricardoquesada:
Thanks.
How is the performance with your changes ?

Panajev:
Well, the number of redundant calls is basically trivial now (compared to before) and performance seems improved, but in order to give you hard numbers I'd need to benchmark it accurately. What do you suggest? Thanks :)

ricardoquesada:
I usually run the performance tests (specially the Performance Sprite
Test   cocos2d-iphone/tests/Performance ) to understand how fast/slow
are the changes.
And I compare the results with these tests:
http://www.cocos2d-iphone.org/wiki/doku.php/perf_test:2_0_0

If you can run it, great. Otherwise, I will run them before the merge.
Thanks.

On 13 February 2012 11:42, Panajev
reply@reply.github.com
wrote:

Panajev:
I am running the tests, Sprites and Particles, and looking at them performance seems to be about the same.One area that would need optimization (NEON) would be the function that checks (using kazmath's built in method) the incoming matrix that the user wants to update: the faster this is done, the better.

Right now, most of the redundant state changes Cocos2D performs are 1.) updating the MVP matrix each frame and 2.) calling glViewport(). Those two account, without the current patch, for the majority of redundant uniform updates. The more uniforms are used by custom shaders the higher the chance that uniform caching helps Cocos2D out.

Still, sorry for not giving you hard numbers, but I am having a bit of  a hard time collecting them when some of them fluctuate on the screen :(.

Panajev:
Example: D(5) with 4,000 particles stays most of the time at 8.6 FPS (from what I read) with infrequent dips below that (blips). On the results page I see ~8.

ricardoquesada:
thanks for the info!
I will try to merge it during this week.

Panajev:
I have added another commit with some further cleanup work (removing some cruft). I hope this is ok.

Panajev:
Do you analyze the output of the performance tests with some tool (and if so, how do you collect the data besides looking at the current FPS counter on screen?) or just see that the number does not produce some nasty results? I find it hard to measure few percentage points of performance whether it is a positive or negative difference :).

ricardoquesada:
For certain tests, I use the FPS counter. For other tests I use the CCProfiler class.
The FPS counter is not very reliable... The "Seconds Per Frame" is a bit more reliable, but not that much.

And usually what I do is just that: See if the numbers are not slower (or not slower than 2% ~ 3%) than the previous version.

Panajev:
Ok, thank you for the explanation :). I agree that most of the times the FPS counter was not reliable, but taking an average FPS value then min and max FPS after running the cycling benchmark could provide more interesting data.

[web [iPhone msg]]

ricardoquesada:
I like your idea. Probably I am going to change a little bit the API, but the idea is good.

ricardoquesada:
thanks. added "uniform" cache here:
https://github.com/cocos2d/cocos2d-iphone/commit/38c97219bfffefd6dca996d70e91db1bfc29a72d

Panajev:
Beautiful, I like the elegance of this implementation :), thank you for
integrating the code and improving on it!
The -updateUniformLocation method does its work more cleanly, but should
also be faster than before. Important, because this was the potential
bottleneck for such a cache.

On Mon, Feb 20, 2012 at 11:49 PM, Ricardo Quesada <
reply@reply.github.com

ricardoquesada:
thanks.
I have not done any performance test between old_cache vs. new_cache,
but I think the new one should be little bit faster since it doesn't
use objective-c data structures...

On 20 February 2012 15:09, Panajev
reply@reply.github.com
wrote:

PR #171, opened 2012-02-23
TouchDispatcher has been finally fixed:) See comments for more details.

SG7:
. [FIX] TouchDispatcher: correctly handles addition and removal of
handlers when being inside a touch handler ( issue #1084  #1139 #1173
#1267)

. [FIX] TouchDispatcher: correctly handles priority change of the
handler when being inside a touch handler
. [NEW] TouchDispatcher: migration from NSMutableArrays to CCArray
along with the change of internal processing allows for substantial
improvements in functions' speed.
 [NEW] TouchDispatcher: user can choose which handlers process the
touches first: Standard or Targeted.
. [NEW] TouchDispatcher: user can choose one of three sorting
algorithms: InsertionSort, QSort, LMerge or none at all.
. [NEW] TouchDispatcher: user can supply his own comparator for
customized sort for the sorting algorithm. ZOrder comparator is
provided. Inverse priority sort is also an option.
. [NEW] TouchDispatcher: Introduction of new dispatcher properties:
'tag', 'disable', 'remove' and new API functions working with ranges of
attribute values give the user the ability to pinpoint and control
handler(s).
. [NEW] TouchDispatcher: New API functions give user full control over
the properties of handlers. They also allow proper disabling and
removing handlers inside and outside touch callbacks.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -6,13 +6,13 @@ http://www.cocos2d-iphone.org/forum
 
 cocos2d logos by Michael Heald (http://www.fullyillustrated.com)
 
-
-
 cocos2d for iPhone authors for v1.1
 -----------------------------------
 
-Lead Developers:
+Lead Developer:
 	Marco Tillemans (http://www.welcometothefabrik.com/)
+
+Former Lead Developer:
 	Ricardo Quesada (http://www.cocos2d-iphone.org)
 
 
@@ -31,6 +31,7 @@ Contributors (alphabetically ordered):
 	Word Wrap works on LabelTTF on Mac
 	bug fixes for failure to compile on 32-bit mac builds
 	created new drawPrimitives functions, ccDrawRect, ccDrawSolidRect and ccDrawSolidPoly.
+    bug fix for CCTMXXMLParser in release mode
 	
 * Jerrod Putman (http://www.tinytimgames.com)
 	Original author of ARC compatibility code
@@ -65,32 +66,49 @@ Contributors (alphabetically ordered):
 
 * Nicky Weber (https://github.com/nickyEnjoysWoogoo):
 	Added convenience function to create a ccColor4F from color components	
+    
+* Nicolas Barrios (http://wefiends.com)
+    Fix for CCArray for ARC compatibility 
 
 * podhrask...@gmail.com
 	TouchDispatcher: correctly handles addition and removal of handlers when being inside a touch handler. patch
 
 * Ricardo Ruiz:
 	Added support for RGB888 textures
+    
+* Rick Smorawski
+    Added full unicode support to CCLabelBMFont
 
+* Samuel J. Grabski http://www.sg7.com
+	Created a new touchdispatcher implementation
+    New touchdispatcher performance tests
+    
 * Scott Lembcke (AKA slembcke) (http://chipmunk-physics.net/)
 	Improved ARC compatiblity code
 
 * Sergey Tikhonov (AKA haqu) (http://haqu.net)
 	Removed warning in CCMenuItem
 
-* slycrel (http://folderapp.blogspot.com/)
-	TMX Map: fixed crash when multiple layers and tiles were used with rotation
-
 * Shakthi:
 	Added CCNode#onExitTransitionDidStart method
 	Fixed fading background music issue
+    
+* Skyhawk (http://www.hkasoftware.com) 
+    Fix for addFrame method in CCAnimation
 
 * slycrel (http://geopherlite.blogspot.com/)
 	Added support for flipped tiles in TMX maps. Also added test case
 	Added support for in-memory TMX map creation. Also added a test case
+    TMX Map: fixed crash when multiple layers and tiles were used with rotation
+
+* Spencer Ho
+    fixed 3d projection settings for ipad3 
 
 * Stepan Generalov (iPsi) (http://www.itraceur.ru/)
 	FileUtils: Loads -hd resources even if the -sd resource is not present. patch
+
+* Steve Kanter
+    new file extension for ipad retina
 
 
 cocos2d for iPhone authors for v1.0
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,30 +1,47 @@
-version 1.1-beta2 xx - 2012
+version 1.1-rc0 .. -..-2012
+. [FIX] CCAnimation: bug in addFrames that resulted in incorrect delayPerUnit (issue #1341)
+. [FIX] CCSprite: CCSprite bug in setDisplayFrameWithAnimationName
+
+version 1.1-beta2b 14-Mar-2012
+. [FIX] CCUtils: missing method declaration
+
+version 1.1-beta2 12-Mar-2012
 . [NEW] v2.0 forward ported changes:
 			LabelBMFont: added setFntFile method
 			Animation: Added new animation format that supports delays per frame and notifications per frame
 			Actions: Added ActionTargeted action. Useful to target other nodes from a sequence.
 . [NEW] All: Code is compatible with ARC (issue #1199) 
+. [NEW] CCArray: new sorting methods, replaceObjectAtIndex, isEqualToArray
+. [NEW] ccFileUtiles: new iPad retina file extension (ipadhd)
 . [NEW] DrawPrimitives: Added new functions, ccDrawRect, ccDrawSolidRect, ccDrawSolidPoly
 . [NEW] DrawPrimitives: Added batch drawing of lines, ccDrawLines (issue #1081)
 . [NEW] Node: Added onExitTransitionDidStart (issue #792)
 . [NEW] ParticleSystem: supports animations in new animation format 
 . [NEW] TileMap: supports in-memory TMX map creation (issue #1239)
 . [NEW] TileMap: Flipped tiles can be changed during runtime (issue #1264)
+. [NEW] Touchdispatcher: new implementation, supporting changing dispatchers based on (ranges) of properties, custom sorting and more (full list in online release notes)
 . [NEW] Types: Added convenience function to create a ccColor4F from color components
 . [FIX] Actions: incorrect behavior of CCRepeat
+. [FIX] CCArray: equalilty is tested by isEqual instead of comparing pointer addresses
+. [FIX] CCArray: CCARRAY_FOREACH macro is now ARC compatible
+. [FIX] CCLabelBMFont: full unicode support
+. [FIX] CCTMXXMLParser: removed unused variables in release mode
 . [FIX] ColorLayer: incorrect size in init method
+. [FIX] Director: 3d projection setting for iPad retina
 . [FIX] Denshion: Background music fades ok (issue #1304)
 . [FIX] EAGLView: always destroy and recreate OpenGL buffers completely (issue #1254) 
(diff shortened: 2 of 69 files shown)


ricardoquesada:
thanks. I will review it during the weekend.

ricardoquesada:
Hey SG7,

Thanks for the pull request. I am reviewing it, and it is somewhat difficult for me to merge your changes because your pull request contains many changes in just one pull request. Also, some of the bug fixes it claims it fixes, were already fixed (eg: issue #1084 and issue #1173 are already fixed).

Would it be possible to split it in different pull request ? (eg: one for issue #1139, another for the priorities, etc.)

If not, could you tell me which parts of the code fixes issues #1139 and #1267 ?

Thanks.

PS: The performance test is a good addition. I will add it to the perf-test suite.

SG7:
> Thanks for the pull request. I am reviewing it
> Thank you for taking time to review it!
> 
> it is somewhat difficult for me to merge your changes
> The best would be to just swap/override old files.
> 
> your pull request contains many changes in just one pull request
> Unfortunately, there was no simple a few liner fixer for all the issues.
> I had to redo the mechanism of adding, removing and changing priorities of
> the handlers to assure proper work of the whole API in the touch callbacks.
> I also migrated to CCArrays to gain speed.
> To add flexibility I have introduced additional API, which gives user full control over the handlers. New API gives additional boost to the speed. User can execute complex block operations according to handler's priority, tag or state.
> 
> Also, some of the bug fixes it claims it fixes, were already fixed (eg: issue #1084 and issue #1173 are already fixed).
> Yes, this pull claims to FIX and cover ALL existing issues with the current
> TouchDispatcher. It also fixes few unreported problems.
> The older proposed patches fix only narrow cases still leaving gaps and
> holes in the code. The main problem is that the old way of handling removal and addition of the handlers falls short to do the proper job when API functions are called from inside touch handler.
1. The patch proposed for issue #1084 does not fix the structural problem
   of CCTouchDispatcher.
   The author of #1084 correctly traces and explains the problem.
   He says: "I get an object that was just destroyed left in the
   TouchDispatcher."
   This is also an issue for #1139 and #1267 - destroyed objects ale left
   dangling in the TouchDispatcher array. The proposed #1084 fix works only
   for very narrow (described in the #1084) particular case but fails apart in
   other scenarios.
2. Proposed fix #1173 also does not cover all cases.
3. Proposed fix in #1267 is the most comprehensive from all proposed
   patches. It is not efficient though, does not cover API for priorities and
   requires two tables to handle removal and addition of handlers.

Again it does not cover all cases.

The most critical thing which I changed is how the removal of the handler
is done. Before (in the locked state), dispatcher had to create an object and add it to the separate array. Than that temporary object had to be removed as well as the handler. I avoid creating temporary objects. Handlers are simply marked for removal. That includes handlers already waiting to be added!
This approach avoids any synchronization issues between add and removal operations. It avoids memory leaks and ghost handlers.
1. One can easily check that old files STILL do not handle callback
   functions properly.
   How to check it? Compile attached performance test with old files:
   CCTouchDispatcher.h CCTouchDispatcher.m CCTouchHandler.h CCTouchHandler.m
   In the PerformanceDispatcher.m
   change line 15 with COMPILE_FOR_NEW_API from 1 to 0

#define  COMPILE_FOR_NEW_API                    0 // USE 0 TO COMPILE WITH 4 OLD FILES FOR OLD API

Compile and touch red label on the screen:
 "Callback - It should not crash."
It will crash with the message:
_DispatcherPerformance[83389:f803] *_\* Terminating app due to uncaught
exception 'NSInternalInconsistencyException'*
If you swap old files with new files everything is fine!
1. >Would it be possible to split it in different pull request ? (eg: one for issue #1139, another for the priorities, etc.)
   I am afraid that would be not simple or practical.

> If not, could you tell me which parts of the code fixes issues #1139 and #1267 ?
> This is the main function properly handling removal and additions of handlers (#1080, #1139, #1173,  #1267):
> -(void) safeProcessing // process messages to the Dispatcher sent by user's touch callbacks
> {
>     // -- remove marked for removal delegates --
>     [self forceRemoveMarkedDelegates];
>     //------------------------------------------
>     CCHandlersToDo *todo;
>     CCARRAY_FOREACH(handlersToDo, todo) {
>         switch (todo.type) {
>             case kCCAddTargetedHandler:
>                 [self forceAddHandler:todo.handler doNotSort:todo.arg type:kCCTargeted];
>                 break;
>             case kCCAddStandardHandler:
>                 [self forceAddHandler:todo.handler doNotSort:todo.arg type:kCCStandard];
>                 break;
>             default:
>                 break;
>         }
>     }
>     [handlersToDo removeAllObjects];
>     //  -- process deleyed requests from user's touch callbacks --
>     [self processCallbackRequestsToTheDispatcher];
>     //------------------------------------------------------------
> }
> In addition the:
> '-(void) processCallbackRequestsToTheDispatcher;'  properly handles priority changes and sorting.
> The main function which handles all critical operations is:
> -(int) alterTouchHandler:(ccDispatcherDelegateType)dType delegate:(id)delegate
>            fieldToSearch:(ccHandlerFieldName)searchField
>                arg1:(int)v1
>                arg2:(int)v2
>                 operator:(ccOperators)op
>             fieldToAlter:(ccHandlerFieldName)fieldToAlter withValue:(int)nV;
> The other important function is:
> - (int) retrieveField:(ccHandlerFieldName)field delegate:(id)delegate type:(ccDispatcherDelegateType)delegateType;
>   Two above function are involved in handling almost all API calls. Most of the API functions are just sugar functions calling directly one of the above functions.
1. The backward compatibility is fully preserved.

This pull delivers full compatibility with old API and adds new API calls.

The main differences are:
a) Touch Dispatcher works properly in the callbacks.
b) API operations are much, much faster.
c) New API gives user full control over the touch handlers.
d) Priorities can be configured according to user wishes. They can be inverted, they can follow z-order, they can be fully customized. They can be dynamically changed or ignored.
e) Attached Dispatcher Performance Test executes test speeds of the dispatcher API. It tests all new API functions. New Dispatcher was also independently tested in a few applications.
f) This pull is not a few line patch or hack. This pull is a result of serious coding effort to deliver much better and stable Touch Dispatcher for cocos2d framework as my thank you token to the cocos2d community.

I welcome all comments and questions.

psineur:
Wow, really big Pull Request. It's hard to me to review it.
That's why atomic commits is so important.

There's one issue in Cocos2D-iPhone-Extensions project, that is created about this: https://github.com/cocos2d/cocos2d-iphone-extensions/issues/97

Thank you for information @SG7

psineur:
I think that new features are very interesting, but i would like to review them commit-by-commit.
Anyway, i'm not currently using some of them, so only CCArray migration affects me, due to that issue in extensions repo.

I set notifications ON on this Pull Request, but if anyone will notify me about when this will be merged - it will be very helpfull.

Thanks!

SG7:
@psineur Thanks for your comment and extensions! 

It looks like only your CCScrollLayer uses private internals of CCTouchDispatcher via category.  I can also let you know when this pull gets merged.

The change of the container to CCArray allowed significant speed improvements and it is in line with the rest of the cocos2d code.

Without running the test suit it is hard to see most of the benefits of the new TouchDispatcher.

This is just one example:

|---------------------------------------------------------------------|
|   Hardware: latest iPod4                                            |
|                                                                     |
|   SPEED TEST:  (5) && (5.c) in the test suite                       |
|                                                                     |
| Changing priorities of the touch handlers                           |
| (50% standard 50% targeted).                                        | 
|---------------------------------------------------------------------|
1. FPS (Frame rate per second)
   |=====================================================================|
   |       Changing priorities   (FPS)                                   |
   |=====================================================================|
   |     NR       |     CURRENT    |             NEW DISPATCHER          |
   |  OF HANDLERS |   DISPATCHER   | Same API fun. | added API function  |
   |=====================================================================|  
   |    # 200 #   |       13.2     |       34\*     |        60  FPS      |
   |---------------------------------------------------------------------|
   |    # 400 #   |       3.7      |       11\*     |        38  FPS      |
   |---------------------------------------------------------------------|
   |    # 600  #  |       1.6      |       5.5\*    |        20  FPS      |
   |---------------------------------------------------------------------|
-  (*) in the touch callbacks (when new dispatcher is used) the FPS increases to the value in the right column
  due to internal optimization. 
  1. Speed (in milliseconds) 
     |=====================================================================|
     |       Changing priorities    (ms)                                   |
     |=====================================================================|
     |     NR       |     CURRENT    |             NEW DISPATCHER          |
     |  OF HANDLERS |   DISPATCHER   | Same API fun.   | added API function|
     |=====================================================================|  
     |   #  200  #  |        55      |       13.6\*     |      4.7  ms      |
     |---------------------------------------------------------------------|
     |   #  400  #  |       212      |       47.7\*     |     16.6  ms      |
     |---------------------------------------------------------------------|
     |   #  600  #  |       485      |      100.5\*     |     31.7  ms      |
     |---------------------------------------------------------------------|
- (*) in the touch callbacks (when new dispathcer is used) the speed increases/time drops to the value
  specified in the right column due to internal optimization.

Edit.
Description got shifted so I put some explanations again here:

The first column represents number of handlers affected. 
Second column holds results for current dispatcher.
Third column is for NEW dispatcher with same api function.
Forth column is for NEW dispatcher with new api function for changing the priority.

Thanks.

araker:
I played with the performance tests and I like it, very powerful new functionality. It seems complex at first, so I think it's good to add some examples for more advanced use cases. I'll also integrate the performance tests.

I'm planning to integrate this during the weekend.

araker:
One thing I doubt that will be useful is the zOrderComparator, it compares the zOrder of the delegate. But when the delegates have different parents, comparing the zOrder is meaningless, since it's not a absolute number, but relative to the parent.

SG7:
Thank you araker for taking close look at the new API!

You are absolutely right about zOrderComparator. It is useful for narrow case when all delegates have same parent.
I do not mind dropping zOrderComparator from the API list.

In practice, user can configure CCTouchDispatcher to use his custom comparator (programmed exactly to his needs) via 'setUsersComparator' method: [[CCTouchDispatcher sharedDispatcher] setUsersComparator:comparator];

araker:
I'll  remove it from the dispatcher code and use it as an example on how to set a user comparator. I'll also place all CCArray extensions in CCArray itself, they're useful for general purpose as well.

SG7:
Very good idea. Thanks!

araker:
I've integrated this code .. https://github.com/cocos2d/cocos2d-iphone/pull/183

I didn't add any more test cases, the' various' test contains an example of all functionality, including the more advanced usages.

SG7:
This implementation solves yet another issue (#1279) with CCTouchDispatcher: 

http://code.google.com/p/cocos2d-iphone/issues/detail?can=2&q=1279&colspec=ID%20Type%20Status%20Priority%20Milestone%20Component%20Owner%20Summary&id=1279

2012-03-24 20:25:52.429 ProblemTest[4116:cb03] **\* Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[NSCFString claimedTouches]: unrecognized selector sent to instance 0x69767e0'

araker:
Thanks, it's fixed..

https://github.com/cocos2d/cocos2d-iphone/commit/9d0f95e982234398597d3debc229adef8d1cda68

PR #174, opened 2012-02-23
Fix Error Handling

cwinkler:
- Change Errorhandling to use Protocols Errorhandling method already implemented, resolving two compiler errors in release mode.

### DIFF
--- a/cocos2d/CCTMXXMLParser.m
+++ b/cocos2d/CCTMXXMLParser.m
@@ -186,7 +186,7 @@
 	[super dealloc];
 }
 
-- (NSError*) parseXMLData:(NSData*)data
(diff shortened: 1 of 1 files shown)


cwinkler:
This was pretty useless, because NSAssert only works in Debug Mode, leaving `err` as an unused variable which led the compiler to fail in release mode.

araker:
Thanks for pointing it out. I fixed it a bit differently. 

https://github.com/araker/cocos2d-iphone/commit/e348e2ab0a817ff38e9b9dd655aa82cd99acabc7.

Leaving it open for 2.0.

araker:
Already fixed in 2.x. Closing

PR #175, opened 2012-02-25
Support for unicode characters > 2048 in CCLabelBMFont

rickms:
This is a follow up to http://www.cocos2d-iphone.org/forum/topic/29494

The current version of CCLabelBMFont, allocates a 2048 sized array of ccBMFontDef for glyphs, this has two issues, one it over allocates memory for 2048 fonts that may not be used, and it does not allow for support of unicode characters above 2048.  I've modified CCLabelBMFont to use a hash to store it's fonts.  I realize this may slow down performance a bit (hash table lookup) but it seemed the best way to support all unicode characters in a CCLabelBMFont

Thanks
Rick

### DIFF
--- a/cocos2d/CCLabelBMFont.h
+++ b/cocos2d/CCLabelBMFont.h
@@ -52,6 +52,8 @@ typedef struct _BMFontDef {
 	int yOffset;
 	//! The amount to move the current position after drawing the character (in pixels)
 	int xAdvance;
+    // Enable Hash Table
+    UT_hash_handle hh;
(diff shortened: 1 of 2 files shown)


ricardoquesada:
Thanks. (already fixed in v2.0 https://github.com/cocos2d/cocos2d-iphone/commit/2c5236b1b4eecaacdd75685023e933fa210295a1 )

But leaving pull request open since it is still valid for v1.1

araker:
thanks, merged.

PR #180, opened 2012-03-08
A Label Batch Node Font Class - CCLabelBNFont

mikezang:
CCLabelBNFont is a subclass of CCSprite

 Features:
    - Can be used with other CCSprites in the same CCSpriteBatchNode
    - Treats each character like a CCSprite. This means that each individual character can be:
    - rotated
    - scaled
    - translated
    - tinted
    - chage the opacity
    - It can be used as part of a menu item.
    - anchorPoint can be used to align the "label"
    - Supports AngelCode text format

 Limitations:
    - All inner characters are using an anchorPoint of (0.5f, 0.5f) and it is not recommend to change it because it might affect the rendering

 CCLabelBNFont implements the protocol CCLabelProtocol, like CCLabel and CCLabelAtlas. CCLabelBNFont has the flexibility of CCLabel, the speed of CCLabelAtlas and all the features of CCSprite. If in doubt, use CCLabelBNFont instead of CCLabelBMFont / CCLabelAtlas / CCLabel.


araker:
Looks good, thanks.

mikezang:
Thanks for your message. By the way, Do I have to commit them to v2.0 branch?

araker:
I don't think that's necessary, Riq can take your commits and apply them on the 2.0 branch as well.

PR #181, opened 2012-03-09
Enable separate iPad Retina graphics

stevekanter:
Using the same method as having "-hd" graphics for retina and "-ipad" graphics for non-retina iPad, this change adds the "-iPadHD" extension.  When an iPad is detected with 2x scale, it uses the "-iPadHD" file, while still maintaining the old file rules.

### DIFF
--- a/cocos2d/Support/CCFileUtils.h
+++ b/cocos2d/Support/CCFileUtils.h
@@ -92,11 +92,27 @@
  */
 +(void) setiPadSuffix:(NSString*)suffix;
 
+
+/** Sets the iPad retina suffix to load resources.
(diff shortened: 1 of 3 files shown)


ricardoquesada:
thanks. applied on v2.0. Leaving pullrequest open.

suffix used: "-ipadhd" instead of "-iPadHD".

araker:
Thanks, fixed in 1.1 as well. Used the same suffix as in 2.0. 

https://github.com/araker/cocos2d-iphone/commit/a28dc0e7c626c1ad7fd2fc2360e5ad4e3047b17d

PR #187, opened 2012-03-13
ARC: Fix pointers to ownership-qualified types in mergesortL

wefiends:
Used the existing CCARRAY_ID typedef to ensure assignments have the same ownership qualifiers, otherwise the new mergesortL and insertionSort methods throw errors with ARC.


araker:
Thanks, you saved me some work. That was on my to-do list for today.

PR #190, opened 2012-03-15
Tiled 0.8 TMX tile rotation support

slycrel:
Hey guys, here is the tile rotation code discussed here:  http://www.cocos2d-iphone.org/forum/topic/27784  I think I might be doing something wrong with my pull request, I can't seem to get the 4 other useless commits out of there?  It's a pretty simple change overall, let me know if I need to do something different here.  Thanks!


ricardoquesada:
Merged into v2.0 with some fixes.
https://github.com/cocos2d/cocos2d-iphone/commit/c6fc8206c9da637065bab3ca0bca78e28533b0f8

araker:
Fixed in 1.1, using the fixes from Riq. 

https://github.com/cocos2d/cocos2d-iphone/commit/bf0b216ad5f258d7db5f08d10feff6ab99ab9e81

PR #191, opened 2012-03-17
Made CCLOG little more verbose.

Shakthi:
Added function name to CCLOG.
To make cclog output something like this.

2012-03-17 11:21:21.604 XXX[5172:207] -[CCDirector init] : cocos2d: cocos2d v1.1.0-pre
Instead of 
2012-03-17 11:21:21.604 XXX[5172:207]  : cocos2d: cocos2d v1.1.0-pre


araker:
Thanks, I like it.

PR #194, opened 2012-03-29
Gles20

kstenerud:
These commits add the following functionality:
- CCLabelTTF now supports vertical alignment in addition to horizontal alignment.
- CCScheduler now differentiates between "system" and "user" tasks (using the special priority kCCPrioritySystem). The default action manager in CCDirector is added as a system task. New methods allow unscheduling and pausing all tasks above a certain priority (so you can affect all non-system tasks by specifying kCCPriorityNonSystemMin, for example).
- CCActionManager now supports pausing/resuming all running actions.

All tests build and run in iOS and Mac.


ricardoquesada:
Thanks. I will do some little API changes and I will merge it.

PR #195, opened 2012-03-29
Make CCParticleSystem unschedule its update selector on dealloc

mhaylock:
We ran into an issue where a CCParticleSystem was being created at the same memory address as a CCParticleSystem that was previously dealloc'd. We would then (with `COCOS2D_DEBUG=1`) get a "CCScheduler: You can't re-schedule an 'update' selector'. Unschedule it first" error.

Adding an unschedule to the CCParticleSystem dealloc fixes this and seems like the correct behavior - it doesn't make sense to leave a selector scheduled for an object that no longer exists?

### DIFF
--- a/cocos2d/CCParticleSystem.m
+++ b/cocos2d/CCParticleSystem.m
@@ -342,6 +342,8 @@
 
 -(void) dealloc
 {
+	[self unscheduleUpdate];
+
 	free( particles );
 
 	if (animationFrameData_) 


djones:
+1

lukelutman:
I ran into a problem with this change causing EXEC_BAD_ACCESS crashes in [CCParticleSystem dealloc]. Here's the bug report: http://code.google.com/p/cocos2d-iphone/issues/detail?id=1405

PR #198, opened 2012-04-05
Make startAnimation and stopAnimation idempotent in CCDirector

kstenerud:
Since CCDirector is a view controller, and is also a singleton, it causes some problems when moving between UIKit and cocos2d. I've made a custom view controller to get around this, but CCDirector still gets view notifications, and shuts itself down (stopAnimation). Thus, I need to be able to safely call startAnimation from my own view controller without worrying about it being active already (and throwing an assert as a result).


SG7:
Nice and important improvement!

PR #199, opened 2012-04-05
Changed %@ = %08X in format strings for printing memory addresses, other format string type issues

kaicherry:
...to %@ = %p. The type 08X is "wrong" and...future...versions of Apple LLVM are very (very) strict about format strings. Without this change compilation fails (LLVM > 3)

Similar changes to types specified in format strings changed, e.g. "Unsupported orientation: %@", orientation_"; orientation is an int. Without this change, compilation fails (LLVM > 3)

### DIFF
--- a/cocos2d/CCAction.m
+++ b/cocos2d/CCAction.m
@@ -62,7 +62,7 @@
 
 -(NSString*) description
 {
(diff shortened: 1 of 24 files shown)


araker:
Thanks

PR #200, opened 2012-04-05
Removed warnings becuase of unrecognised arc related pragma 

Shakthi:
Hi,

I am using xcode3.2.6(LLVM compiler 1.7),  iOS-sdk 4.3   on snow leopard. 

The Compiler doesn't recognize -Warc-performSelector-leaks and -Warc-non-pod-memaccess and gives a bunch of warnings for for 'ccCArray.h'.
That results  54 errors in the release build.

I just wrapped them in "#if defined(__has_feature) && __has_feature(objc_arc)"

Thanks


araker:
Thanks, I'm using LLVM 2.0 myself, so I can't test this anymore.

PR #206, opened 2012-04-30
Add sound events to CocosDenshion's SimpleAudioEngine.

desplesda:
Hi there!

I added support for sound events in CocosDenshion, which allow you to have pretty sophisticated control over sound playback while minimising the amount of code a game needs to write. It also makes it easier for sound designers to tweak the sounds used in the game without having to rebuild the app. Finally, it has support for queueing voiceover lines, to prevent collisions. This based on a modified version of CocosDenshion that we're using in a very audio-heavy iPad app.

Full documentation follows (and this file is also included in the commit.) Let me know what you think!
# CocosDenshion Sound Events

Version 1.0
30 April 2012

© Secret Lab: http://www.tothesecretlab.com
Coded and written up by Jon Manning.

Questions? Comments? jon@secretlab.com.au
## Introduction

So we happened to be working on a project that involved quite a lot of sound effects, including looped background audio, sound effects and voiceover content. CocosDenshion was our natural choice for a simple-to-use sound system.

However, our audio designers didn't want to have to deal with the pain of rebuilding the application every time they wanted to change a file, and didn't want the app to hard-code the sound file names, since they often named the sound files after what the sounds contained and not when the sound was used. They also wanted to indicate which files should be looped, tweak volume and pitch, and other little things.

We also wanted to have good support for voiceover lines, which have special requirements in that you can't ever have two voiceover lines play at the same time, since they collide badly. So, we added a queue system that gives you better control over how to deal with voiceover lines.

So, we came up with this: a "sound event" system, which separates the audio from the context in which it's used. The way it works is that you create a JSON file that maps sound event names (like "player died") to sound files (like "scream.wav"). If you want to get fancy, you can provide settings that control additional settings.

Finally, the whole thing is set up to allow you to modify both the sound event mapping and the sound files themselves through iTunes File Sharing. You can hand copies of your game to your audio designers, and they can replace and tune files by updating the files stored in your app, without having to get new builds.
## The sound events list

The sound events list is a JSON file called `SoundEvents.json`. You add it to your project, so that it's compiled into the project. If the app finds a file called `SoundEvents.json` in the Documents directory in its sandbox, it will use that file instead.

The sound events list looks like this:

```
{
    "player-died": "scream",
    "shot": "rifle-shot",
    "level-intro": {
        "file":"voiceover-levelintro",
        "queue":"always"
    },
    "level-warning": {
        "file":"voiceover-watchout",
        "queue":"never"
    },
    "level-background-music": {
        "file": "levelmusic03",
        "background": true,
        "loop": true
    },
    "explosion" : {
        "file":"explode",
        "gain":0.2,
        "pitch-variability": 0.1
    }
}
```

Each entry in the list can be either a string or a dictionary.

If the entry is a string, the item's name is the sound event, and the item's value is the name of the audio file. So, for the first entry, we're creating a sound event called "player-died", which will play the file "scream.wav" when triggered. (If you don't include ".wav" in the audio file name, CocosDenshion will add it for you.)

If the entry is a dictionary, you can provide additional information about the sound event. The settings you can provide are:
- `file`: String. The name of the audio file that should be played when the sound event is triggered. You must include this setting, or the event won't play a sond.
- `background`: Boolean. If true, the audio file will play in the background. Only one background audio file may play at once; if you play one when another is currently playing, the first sound will stop.
- `loop`: Boolean. If true and the file is a background sound, the sound will loop until stopped. This setting is ignored if the sound is not a background sound.
- `gain`: Number. Setting this value to 0 makes the sound silent, and setting it to 1 (the default) makes the sound play at full volume. 
- `pitch-variability`: Number. Setting this number to anything greater than 0 will make the sound play at a slightly varied pitch (making it great for things like collision noises or anything you need to play over and over without getting too repetitive). Higher numbers mean more variability.
- `queue`: String, either "always" or "never". This setting controls whether the sound works with the voiceover queue (see below.)

Note that SoundEvents.json must be valid JSON, or else it won't load. A great OS X JSON editor is "Jason": http://olivierlabs.com/jason/
## Working with sound events

To trigger a sound event, you do this:

```
[[SimpleAudioEngine sharedEngine] playSoundForEvent:@"player-died"];
```

This makes the sound engine play the sound associated with the provided sound event. The same call is used for background audio and voiceovers.

When you're done with a scene, you do this:

```
[[SimpleAudioEngine sharedEngine] stopSounds];
```

This stops the background music, and cancels any queued voiceover lines.

That's the entire API - we realised that you want to keep the amount of audio code in the game to an absolute minimum, so the entire thing is based around triggering audio events.
### The voiceover queue

Voiceover lines are somewhat special cases, since if you play two at a time, they collide and you can't understand them. To solve this problem, we added queued audio support.

The philosophy of the system is this: voiceover lines should never overlap, and you shouldn't end up with a situation where repeatedly triggering a voiceover event queues a huge amount of lines up.

In the sound events system, you can mark sound events should participate in the queue. To do this, you provide a `queue` setting in the sound event's dictionary. Here's an example of two (from above):

```
"level-intro": {
    "file":"voiceover-levelintro",
    "queue":"always"
},
"level-warning": {
    "file":"voiceover-watchout",
    "queue":"never"
}
```

If you set the `queue` setting to "always", the sound file will be added to the queue. If the queue is empty, the sound file will start playing at once. If the queue is not empty, the sound file will wait until it's ready. Multiple items can be added to the queue at once.

If you set the `queue` setting to "never", the sound file will _only_ play if the queue is empty. If a sound file that's in the queue is currently playing, the sound won't play. 

If you don't provide a `queue` setting, the sound event ignores the queue and plays, overlapping any existing sounds.

This allows you to indicate which voiceover lines are important, and should always be played, and which voiceover lines are less important and can afford to not be heard.

Mad props: This queue system was directly inspired by Darren Korb (@DarrenKorb), from Supergiant Games.

### DIFF
--- a/CocosDenshion/CocosDenshion/SimpleAudioEngine+SoundEvents.h
+++ b/CocosDenshion/CocosDenshion/SimpleAudioEngine+SoundEvents.h
@@ -1,0 +1,35 @@
+//
+//  SimpleAudioEngine+SoundEvents.h
+//
+//  Created by Jon Manning on 29/02/12.
+//  Copyright (c) 2012 Secret Lab. All rights reserved.
+//
+//  Permission is hereby granted, free of charge, to any person obtaining a copy
+//  of this software and associated documentation files (the "Software"), to deal
+//  in the Software without restriction, including without limitation the rights
+//  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+//  copies of the Software, and to permit persons to whom the Software is
+//  furnished to do so, subject to the following conditions:
+//  
+//  The above copyright notice and this permission notice shall be included in
+//  all copies or substantial portions of the Software.
+//  
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+
+#import "SimpleAudioEngine.h"
+
+@interface SimpleAudioEngine (SoundEvents)
+
+// Play a sound with a given name. Returns YES if the sound is or will be playing, NO if the sound couldn't be played or was blocked by other sound. 
+- (BOOL) playSoundForEvent:(NSString*)eventName;
+
+// Stop the background music and clear the voiceover queue
+- (void) stopSounds; 
+
+@end
--- a/CocosDenshion/CocosDenshion/SimpleAudioEngine+SoundEvents.m
+++ b/CocosDenshion/CocosDenshion/SimpleAudioEngine+SoundEvents.m
@@ -1,0 +1,261 @@
+//
+//  SimpleAudioEngine+SoundEvents.m
+//  ClockPhysics
+//
+//  Created by Jon Manning on 29/02/12.
+//  Copyright (c) 2012 Secret Lab. All rights reserved.
+//
+//  Permission is hereby granted, free of charge, to any person obtaining a copy
+//  of this software and associated documentation files (the "Software"), to deal
+//  in the Software without restriction, including without limitation the rights
+//  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+//  copies of the Software, and to permit persons to whom the Software is
+//  furnished to do so, subject to the following conditions:
+//  
+//  The above copyright notice and this permission notice shall be included in
+//  all copies or substantial portions of the Software.
+//  
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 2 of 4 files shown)


ghost:
It looks good.  Are you willing to support it?  If you don't want to provide support then I suggest you just announce it in the CocosDenshion forum and people can merge it as they see fit.  If you are happy to provide support then I'm happy for it to be included.

desplesda:
Yes, I'm willing to support it for the foreseeable future. This is code that we're using in an ongoing project, so I've got good reason to keep it updated.

I'm not sure if you want to keep it as a category, though.

On 01/05/2012, at 12:55 PM, steveoldmeadow wrote:

> It looks good.  Are you willing to support it?  If you don't want to provide support then I suggest you just announce it in the CocosDenshion forum and people can merge it as they see fit.  If you are happy to provide support then I'm happy for it to be included.
> 
> ---
> 
> Reply to this email directly or view it on GitHub:
> https://github.com/cocos2d/cocos2d-iphone/pull/206#issuecomment-5432780

---

jon manning

secret lab
co-founder

ghost:
It is up to you.  I don't use CocosDenshion any more so I'm happy to hand over the reins to people that use it.  My only concern is that changes are backwards compatible so it doesn't break existing code.

desplesda:
Fair enough!

desplesda:
Hi! Is this still accepted, and do you know when you plan to merge it?

ghost:
I'm not sure when it will be merged, I'm not involved with that.  You need to speak to araker for cocos2d 1.1 or riq for cocos2d 2.0.

desplesda:
Ok! Do you think this would be a better fit for 1.1 or 2.0?

ghost:
Probably both.  You should also start a thread in the CocosDenshion forum and explain what your patch does.  Otherwise I doubt anyone will use it if it is simply merged into the code base.

desplesda:
Will do.

Panajev:
Any update?

ghost:
It is in the queue.  Riq has already stated he has a tonne of stuff to merge in.  At the moment he is concentrating on the javascript bindings.

Panajev:
Thank you for your comment, more than enough explanation for me :). I just wanted to know if it was still queued to happen.

rsanchezsaez:
This looks really great. My support for this being merged into the main tree.

steveoldmeadow: What do you use now instead of CocosDeshion?

ghost:
@rsanchezsaez - I use fmod or Bass.

rsanchezsaez:
Cool, thanks for the tips.

ricardoquesada:
@steveoldmeadow If you are ok with it, I will merge it in develop-v2.
Could someone provide a pull request against develop-v2 branch ? thanks.

ghost:
@riq - It is fine by me.

ricardoquesada:
@steveoldmeadow thanks. I will try to merge next week.

mtorrance:
Hi All,

Was this merged into the 1.1 rc0?

Thanks

ricardoquesada:
No, but I will try to merge them for v2.1

ricardoquesada:
I started merging it for v2.1, but I noticed it is using `NSJSONSerialization` which is iOS >= 5.0.
I will change for .plist support in order to support iOS 4.x

PR #208, opened 2012-05-24
Fixed a bug in CCTexture2D convert to RGB5A1 function

KidstamaticApps:
When you convert RGBA8888 texture to RGB5A1 texture and then render it on black background, you'll see a "ghost" image as if the texture is still RGBA8888. 

 On background lighter than the pixel color this effect disappers.
 This happens because the old convertion function doesn't premultiply old RGB with new A.

 As Result = sourceRGB + destination*(1-source A), then
 if Destination = 0000, then Result = source.

 Here comes the ghost!

 We need to check new alpha value first (it may be 1 or 0) and depending on it whether convert RGB values or just set pixel to 0

### DIFF
--- a/cocos2d/CCTexture2D.m
+++ b/cocos2d/CCTexture2D.m
@@ -420,17 +420,28 @@ static CCTexture2DPixelFormat defaultAlphaPixelFormat_ = kCCTexture2DPixelFormat
 	}
 	else if (pixelFormat == kCCTexture2DPixelFormat_RGB5A1) {
 		//Convert "RRRRRRRRRGGGGGGGGBBBBBBBBAAAAAAAA" to "RRRRRGGGGGBBBBBA"
+		/*
+		 Here was a bug.
+		 When you convert RGBA8888 texture to RGB5A1 texture and then render it on black background, you'll see a "ghost" image as if the texture is still RGBA8888. 
+		 On background lighter than the pixel color this effect disappers.
+		 This happens because the old convertion function doesn't premultiply old RGB with new A.
+		 As Result = sourceRGB + destination*(1-source A), then
+		 if Destination = 0000, then Result = source. Here comes the ghost!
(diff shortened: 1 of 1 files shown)


cocos2d/CCTexture2D.m
```diff
@@ -420,17 +420,28 @@ - (id) initWithCGImage:(CGImageRef)cgImage
...
 		outPixel16 = (unsigned short*)tempData;
 		for(unsigned int i = 0; i < textureWidth * textureHeight; ++i, ++inPixel32)
-			*outPixel16++ =
-			((((*inPixel32 >> 0) & 0xFF) >> 3) << 11) | // R
-			((((*inPixel32 >> 8) & 0xFF) >> 3) << 6) | // G
-			((((*inPixel32 >> 16) & 0xFF) >> 3) << 1) | // B
-			((((*inPixel32 >> 24) & 0xFF) >> 7) << 0); // A
-
-
+			if (((((*inPixel32 >> 24) & 0xFF) >> 7) << 0)) // A can be 1 or 0
```
KidstamaticApps, line 436: Better:
if ((*inPixel32 >> 31))
Sorry, don't want to open a new pull request

ricardoquesada, line 436: Ok. Thanks.

You can push a new commit on your gles20 branch, and that commit will appear in the queue of commits to merge. No need to open a new pull request

ricardoquesada:
Also, if possible, could you send me a test case for this patch ? thanks.

KidstamaticApps:
As I see, you have already merged this. Do you still need a test case? If yes, what is the best way to send it (i have a Cocos2D project for this)? I'm a newbie at github and didn't get all its features yet.
25.05.2012, в 9:13, Ricardo Quesada написал(а):

ricardoquesada:
Yes, if you could send me a simple test case in a project would be great.
ricardoquesada AT gmail

I tested it with some textures and I saw the ghost as well, but I want to make sure that my test case is similar to yours.

PR #209, opened 2012-06-07
Offset on touch position when using overlayed UIScrollView.

Zakay:
When using with additional overlayed views (such as UIScrollView) we get an offset, and this fixes that.


araker:
Thanks, fixed this already..

https://github.com/cocos2d/cocos2d-iphone/commit/f3bdf7d2048ab0deee92f13f6b83342fc6d6a90e

PR #212, opened 2012-06-15
CCGLProgram guards, fixes and debugging info

iomac:
Added some asserts to help with debugging shaders, fixed a couple of
bugs that were hiding/overwriting programLog information (complicating
debugging), and subsequently found/fixed a minor bug in the
CCShaderCache default shader initialization routine.

With these changes it is much easier to figure out why a shader is not
compiling/linking as expected. It also makes it less likely that cocos
will start spewing gl error codes from weird locations because the shader
asserts will break much closer to the actual problem.

### DIFF
--- a/cocos2d/CCGLProgram.h
+++ b/cocos2d/CCGLProgram.h
@@ -139,6 +139,9 @@ struct _hashUniformEntry;
 /** will update the MVP matrix on the MVP uniform if it is different than the previous call for this same shader program. */
 -(void) setUniformForModelViewProjectionMatrix;
 
+/** calls glGetUniformLocation only if the shader program has been properly initialized **/
+- (GLint)uniformLocationForName:(NSString*)name;
+
 /** returns the vertexShader error log */
 - (NSString *)vertexShaderLog;
 
--- a/cocos2d/CCGLProgram.m
+++ b/cocos2d/CCGLProgram.m
@@ -105,9 +105,11 @@ typedef void (*GLLogFunction) (GLuint program,
 
 - (id)initWithVertexShaderFilename:(NSString *)vShaderFilename fragmentShaderFilename:(NSString *)fShaderFilename
 {
-	
 	const GLchar * vertexSource = (GLchar*) [[NSString stringWithContentsOfFile:[[CCFileUtils sharedFileUtils] fullPathFromRelativePath:vShaderFilename] encoding:NSUTF8StringEncoding error:nil] UTF8String];
 	const GLchar * fragmentSource = (GLchar*) [[NSString stringWithContentsOfFile:[[CCFileUtils sharedFileUtils] fullPathFromRelativePath:fShaderFilename] encoding:NSUTF8StringEncoding error:nil] UTF8String];
+    
+    NSAssert1( !vShaderFilename || (vShaderFilename && vertexSource),    @"vertex shader:   %@ - file not found", vShaderFilename );
+    NSAssert1( !fShaderFilename || (fShaderFilename && fragmentSource),  @"fragment shader: %@ - file not found", fShaderFilename );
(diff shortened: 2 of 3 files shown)


Panajev:
In Debug mode I am getting crashes when trying to validate the program during the link phase. I do not think this approach is working well. I kep all your other changes to CCGLProgram and all the changes to ccShaderCache, but I have to revert to the older link method (I left your assert on top, checking if program_ != 0 or not). We should review the method to see what happened (testing on iOS 6 btw). What I am thinking is that the code fails validation and deletes the shader program...

Panajev:
After further analysis I think we should remove checks for program validation status:

``` objective-c
        glValidateProgram(program_);
        glGetProgramiv(program_, GL_VALIDATE_STATUS, &status);
```

apparently validation is run at a point which will produce an error

"cocos2d: ERROR: Failed to validate program: 1 - Validation Failed: Current draw framebuffer is invalid."

then, in our code the program is deleted and everything breaks. apart in the rest of the GL code.

The reason is that this code is executed before the EAGL surface is ready and Cocos2D has started animating:

``` objective-c

2012-08-31 20:30:53.362 wd[22173:c07] cocos2d: cocos2d v2.0.0
2012-08-31 20:30:53.363 wd[22173:c07] cocos2d: Using Director Type:CCDirectorDisplayLink
2012-08-31 20:30:53.475 wd[22173:c07] cocos2d: ERROR: Failed to validate program: 1 - Validation Failed: Current draw framebuffer is invalid.
2012-08-31 20:30:53.477 wd[22173:c07] cocos2d: ERROR: Failed to validate program: 4 - Validation Failed: Current draw framebuffer is invalid.
2012-08-31 20:30:53.478 wd[22173:c07] cocos2d: ERROR: Failed to validate program: 7 - Validation Failed: Current draw framebuffer is invalid.
2012-08-31 20:30:53.479 wd[22173:c07] cocos2d: ERROR: Failed to validate program: 10 - Validation Failed: Current draw framebuffer is invalid.
2012-08-31 20:30:53.480 wd[22173:c07] cocos2d: ERROR: Failed to validate program: 13 - Validation Failed: Current draw framebuffer is invalid.
2012-08-31 20:30:53.482 wd[22173:c07] cocos2d: ERROR: Failed to validate program: 16 - Validation Failed: Current draw framebuffer is invalid.
2012-08-31 20:30:53.483 wd[22173:c07] cocos2d: ERROR: Failed to validate program: 19 - Validation Failed: Current draw framebuffer is invalid.
2012-08-31 20:30:54.096 wd[22173:c07] cocos2d: animation started with frame interval: 60.00
2012-08-31 20:30:54.097 wd[22173:c07] cocos2d: surface size: 2048x1536

```

Panajev:
Replaced the previous code with this:

``` objective-c

- (BOOL)link
{
    NSAssert(program_ != 0, @"Cannot link invalid program");

    GLint status = GL_TRUE;
    glLinkProgram(program_);

    if (vertShader_) {
        glDeleteShader(vertShader_);
    }
    if (fragShader_) {
        glDeleteShader(fragShader_);
    }
    vertShader_ = fragShader_ = 0;

#if DEBUG
    glGetProgramiv(program_, GL_LINK_STATUS, &status);
    NSString* log = self.programLog;

    if (status == GL_FALSE) {
        NSLog(@"cocos2d: ERROR: Failed to link program: %i - %@", program_, log);
        ccGLDeleteProgram( program_ );
        program_ = 0;
    }
#endif

    return status == GL_TRUE;
}

```

ricardoquesada:
Thank you.
I merged a very similar patch from @slembcke. It aborts when the shader fails to compile, and reports the errors.
But I have added your new link code.

PR #213, opened 2012-06-15
old implementation of ccpAngle is imprecise

ghost:
due to prior normalization calculated angle can be quite inaccurate

### DIFF
--- a/cocos2d/Support/CGPointExtension.m
+++ b/cocos2d/Support/CGPointExtension.m
@@ -190,7 +190,8 @@ BOOL ccpLineIntersect(CGPoint A, CGPoint B,
 
 float ccpAngle(CGPoint a, CGPoint b)
 {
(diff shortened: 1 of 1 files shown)


cocos2d/Support/CGPointExtension.m
```diff
@@ -190,7 +190,8 @@ BOOL ccpLineIntersect(CGPoint A, CGPoint B,
 
 float ccpAngle(CGPoint a, CGPoint b)
 {
-	float angle = acosf(ccpDot(ccpNormalize(a), ccpNormalize(b)));
```
ghost, line 193: It is really a nice and useful piece of info. I am glad that you shared this helpful information with us. Please stay us up to date like this. Thank you for sharing.
Calvin Klein Calzoncillos http://www.margen.com/mini/klein/calvin-klein-calzoncillos.html

PR #215, opened 2012-06-23
Fixes uniqueIdentifier warnings in CocosLive using OpenUDID

JoseExposito:
https://github.com/ylechelle/OpenUDID


araker:
CocosLive will not be supported anymore in 1.1 RC0 (2.x. already doesn't support it). I'll leave this patch open so people can implement this themselves.

PR #216, opened 2012-06-29
Removed forward declarations for tFontDefHashElement and tKerningHashElement

Panajev:


Ricardo1980:
I tested the last modification and doesn't work properly. In landscape, my sprite is not centered when first scene is loaded, but centered if I move those 2 lines to their original position.
And BTW, with that modification it seems there is no flicker.

Panajev:
That is strange. So it's either flicker and correct centering or no flicker and wrong sprite positioning at start?
Riq? Ideas?

Ricardo1980:
Did you see that strange behavior or would you like I create a test case?

Panajev:
If you could create a text case I'd appreciate it a lot, thanks :). As far as the missing structs issue, yes I have.

[web [iPhone msg]]

ricardoquesada:
Strange... I cannot reproduce this issue...
I have a test (bug350), and it is working ok.

Ricardo1980:
I created a test case for you. 
Donwload this file.
http://www.4shared.com/zip/m-MpBbM_/TestFlickering.html
If you have problems downloading it, tell me about it.

Now you can see a flicker. If you move those 2 lines after pushScene, there is no flicker, but first sprite is not centered, and what is even more strange, if you change to another scene and come back, position is right.

Test it and tell me.
Thanks for help.

ricardoquesada:
thanks. merged in develop-v2 branch

PR #217, opened 2012-07-01
Stella 1.1

stellasdk:
Adding Cocos2D support for StellaSDK 1A2300

StellaSDK can be used to develop cocos2d games for Android

More information see yeecco.com


stellasdk:
compiler flags __STELLA_VERSION_MAX_ALLOWED/__IPHONE_VERSION_MAX_ALLOWED is used to distinguish between iOS platform and StellaSDK platforms (Android simulator/devices)

PR #218, opened 2012-07-02
Setting the delay per unit in CCAnimation

killerham:
- now modifies the duration of the animation


araker:
Thanks, fixed.. 

https://github.com/cocos2d/cocos2d-iphone/commit/68deedc9fa124132fdddb9e00034f2537868b6ee

PR #219, opened 2012-07-17
Remove the kazmath neon patch.

fables-tales:
This file isn't being used anywhere in the project, it
should therefore be removed from the externals folder.


fables-tales:
yo @cocos2d can you take a look at this?

PR #220, opened 2012-07-17
Fixed block copying issue.

yosit:
This fixes a crash with the block copy

### DIFF
--- a/CocosDenshion/CocosDenshion/CDAudioManager.m
+++ b/CocosDenshion/CocosDenshion/CDAudioManager.m
@@ -262,7 +262,7 @@ static BOOL configured = FALSE;
 		return NO;
 	}
 }
-
+ 
 // Init
 + (CDAudioManager *) sharedManager
 {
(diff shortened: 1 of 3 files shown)


yosit:
from: http://stackoverflow.com/questions/7069174/objective-c-block-property-defined-as-copy-not-working

The underlying issue is libSystem library linked into the app.

The fix is to remove "-weak_library /usr/lib/libSystem.B.dylib" from the target's Linker Flags and replace it with "-weak-lSystem".

ricardoquesada:
What is the best way to reproduce the crash ? I want to add a test case. Thanks.

yosit:
Put the old flag in the linker flags like I mentioned on my last comment

## Sent from my iPhone

## 

Yosi Taguri | Co-Founder | Labgoo | iPhone: http://ahhhpah.com | twitter:
http://twitter.com/yosit | Mobile: +972-54-4816020 | Google Voice US
Mobile: +1 (415) 935-1126 | Email: yosi@taguri.com

On 18 ביול 2012, at 22:15, Ricardo Quesada <
reply@reply.github.com>
wrote:

What is the best way to reproduce the crash ? I want to add a test case.
Thanks.

---

PR #224, opened 2012-07-27
CCDirector startup flicker

cjhanson:
CCDirector.m

runWithScene:
- Added assert to prevent calling this with a scene already on the scene stack
- Removed call to startAnimation
- Added forced draw after pushing the scene

replaceScene:
- Added assert to prevent calling this with the first scene and a note to use runWithScene: instead

pushScene:
- No changes, but I could not add an assert to prevent calling this for the first scene since it is internally called by runWithScene: Maybe it should output a log warning?

CCGLView.m

layoutSubviews
- Added a conditional check to make sure there is actually a runningScene before drawing to prevent drawing the clear color as the first draw

### DIFF
--- a/cocos2d/CCDirector.m
+++ b/cocos2d/CCDirector.m
@@ -366,13 +366,17 @@ static CCDirector *_sharedDirector = nil;
 - (void)runWithScene:(CCScene*) scene
 {
 	NSAssert( scene != nil, @"Argument must be non-nil");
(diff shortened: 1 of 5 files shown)


ricardoquesada:
Sorry, I didn't have the time to look at it during the weekend. I will try to look this Wednesday or Thursday.

PR #225, opened 2012-07-27
Don't allocate the stats labels if setting view to nil

jdmunro:
This creates unused objects when cleaning up the director. We use cocos2d in an app that also uses UIKit. So we frequently initialise/cleanup the CCDirector. During its cleanup, it sets the view to nil and this was allocating labels that can never be used.

You can confirm this using the Allocations Instrument. Do a heapshot before initialising cocos2d, then do one after cleanup. Repeat a few times - you'll see various CC objects on the heap after cleanup has complete.


ricardoquesada:
thanks. fixed.
https://github.com/cocos2d/cocos2d-iphone/commit/39259010fd9912f54f4dcabf438b0e4763fb2315

PR #226, opened 2012-07-27
Ported kaicherry's patch that fixed LLVM > 3 compile errors in v2.x to v1.x

colinmccune:
As of xcode 4.4 / LLVM 4 cocos2d v1 fails to compile due to incorrect format paramaters in strings. Kaicherry committed a patch that fixes these issues in v2.x however these changes were never applied to v1.x. This patch contains the require changes to make v1.x compile successfully.

Original Patch: https://github.com/cocos2d/cocos2d-iphone/commit/f09b1502651765401ffad12d2baf994c909d6c43

### DIFF
--- a/cocos2d/CCAction.m
+++ b/cocos2d/CCAction.m
@@ -64,7 +64,7 @@
 
 -(NSString*) description
 {
-	return [NSString stringWithFormat:@"<%@ = %08X | Tag = %i>", [self class], self, tag_];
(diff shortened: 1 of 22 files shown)


zbyhoo:
+1

araker:
Thanks, applied it in 1.1. 

https://github.com/cocos2d/cocos2d-iphone/commit/27f411a2f5ce6c0c5085fa7cc1629ccde3a97be8

PR #227, opened 2012-07-29
[FIX] CCMenuItemToggle - Tags of children were modified by the component

yannickl:
The same as the pull request #223 but concerning the Cocos2D 1.1.0
https://github.com/cocos2d/cocos2d-iphone/pull/223


araker:
Thanks, fixed.. 

https://github.com/cocos2d/cocos2d-iphone/commit/59d3c4d03df7ba75d6b084999e5fe7e11c85cf78

PR #228, opened 2012-08-09
Fixes crash on Bitmap font when characters are out of range of font

cjhanson:
http://code.google.com/p/cocos2d-iphone/issues/detail?id=1404


ricardoquesada:
thanks. merged in develop-v2

PR #230, opened 2012-08-16
Support for transitions when pushing and popping scenes.

Panajev:
Added support for scene pushing and popping using animated transitions.


rsanchezsaez:
Does this work when popping the last scene? (Transition to black). If it does, this could be really great!

Panajev:
Theoretically, Cocos2D transitions are used to go from one scene to another
scene. It would work well if you had an empty scene for it to transition to
and you used a transition that allowed you to fade with the appropriate
color.

Otherwise, this patch follows the same limitations for pushing scenes and
popping scenes the director previously had. I'll give this some more
thought. I was planning this feature for things like menus and other layers
you might want to pop in and out quickly but with a nice transition effect.
I am also planning to use something other than strings for selecting the
transitions... the current way is not exactly code completion friendly.

rsanchezsaez:
Thanks for the info. Looks quite useful anyway!

Panajev:
Thank you :).

PR #231, opened 2012-08-20
Rendering Target Node - Rendering to texture made simpler.

krzysztofzablocki:
```
Problem
```

CCRenderTexture is abused by many people, people get confused how to render properly into it, like overriding visit methods, mostly because not all cocos nodes are easy to use with CCRenderTexture:
-  CCSpriteBatchNode cannot be visited if it's not inside scene tree, if you add it to scene tree and then call [batch visit] between begin and end of RenderTexture rendering you are rendering twice. That's just wrong. You could 'fix' it by hiding the batch at start and then make it visible just before visiting it inside your renderTexture redraw. I still believe that's far from nice solution. 
- Animations are _paused if nodes are not in scene node, again could be 'fixed' as in previous point.

'fixed' - means workaround and in my book is really poor.

```
Solution
```

Added CCRenderTargetNode - a node that simplifies rendering into CCRenderTexture. 
- Whatever you add into it, will be rendered into CCRenderTexture and only there. 
- Animations will work since nodes will be in scene tree, batches will work as well. No hacking required and only 1 rendering.
- CCRenderTexture has lazy created .renderTargetNode property that can be used for adding renderTexture content nodes into it, they will be rendered into RenderTexture and only there.
- ( Also it could be used to render into CCRenderTexture that's not in scene tree ( if you'd need that kind of functionality ), you can plainly create it with RenderTexture but you need to retain RenderTexture by yourself so that it doesn't get released before this node. )

### DIFF
--- a/cocos2d-ios.xcodeproj/project.pbxproj
+++ b/cocos2d-ios.xcodeproj/project.pbxproj
@@ -1176,6 +1176,8 @@
 		50FD7B3C0FFAA8AC004073B5 /* grossini_dance_atlas.png in Resources */ = {isa = PBXBuildFile; fileRef = 509A7A970F61A2400032F449 /* grossini_dance_atlas.png */; };
 		50FD7B3F0FFAA8BC004073B5 /* background.png in Resources */ = {isa = PBXBuildFile; fileRef = 50C1959C0EE4B90800829067 /* background.png */; };
 		50FD7B450FFAA8C6004073B5 /* atlastest.png in Resources */ = {isa = PBXBuildFile; fileRef = 506882680E957C4A00F943E5 /* atlastest.png */; };
+		6D0E1183C5BE3CB5DA10D03D /* CCRenderTargetNode.h in Headers */ = {isa = PBXBuildFile; fileRef = 6D0E11D552E8A1F6205D918E /* CCRenderTargetNode.h */; };
+		6D0E1D88B980A0D8C05324E6 /* CCRenderTargetNode.m in Sources */ = {isa = PBXBuildFile; fileRef = 6D0E193AA35301C580F6B274 /* CCRenderTargetNode.m */; };
 		6F305B5F0FB9D2790052E700 /* TransformUtils.h in Headers */ = {isa = PBXBuildFile; fileRef = 6F305B5D0FB9D2790052E700 /* TransformUtils.h */; };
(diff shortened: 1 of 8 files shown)


Panajev:
It looks really interesting. It does remove a few roadblocks to CCRenderTexture.

ricardoquesada:
I like. Scheduled for v2.1

ricardoquesada:
I sort of merged it it.
What I did was to add the `CCTargetRenderNode` behavior in `CCRenderTexture`.
For some reason, it is not working properly now... working on the fix.

krzysztofzablocki:
It probably doesn't work because you should call visit on the children ( in renderTexture visit method ) when auto-draw mode is turned off.

I designed it this way ( as separate target node ) because it allows you to have more control over it, yet it's still very simple to use.
- You could draw to offscreen render texture and make effects on it very easily.
- You could add an overlay on top of processed texture content (as a child of render texture) and have it drawn standard way, yet still have some fancy effects in renderTexture content.
  etc.

ricardoquesada:
Thanks. 
Yes, I am calling `visit` on the children. The bug is that for some reason the transformations applied to  `RenderTexture` are not inherited by its children... I found an ugly workaround, but I am trying to figure out the cause of the bug...

`CCRenderTargetNode` is easy to use, and I like it. What I don't like is to have 3 different nodes with similar features:
-  `CCGrabber`:
-  `CCRenderTexture`
-  `CCRenderTargetNode`

Which one should you use and when ? It is not clear for the user. I think it is confusing. That's why I merged `CCRenderTexture` and `CCRenderTargetNode`.

Panajev:
Would the other patch related to viewport and model view transforms in begin and end do any good to the bug you see now @riq?

[web [iPhone msg]]

krzysztofzablocki:
@ricardoquesada there is still a mistake because currently code in repository is calling [children visit] in draw method but ONLY if you have autodraw enabled. You still need to visit children if you don't have autodraw enabled ( default ) or you break backward compatibility

ricardoquesada:
@panajev: Mmm... yes, perhaps.

@krzysztofzablocki: Thanks. What I am doing instead is calling `[sprite visit]`, which works in 99% percent of the cases. But as you mentioned, `CCRenderTexture` could have other children as well

Panajev:
@ricardoquesada it might not solve the issue, that I do not know for sure,
but I think it should at least provide an even cleaner foundation to
explore the issue you are experiencing.

PR #232, opened 2012-08-20
Optimization - eliminated costly [NSNumber floatValue] calls

thomas-brx:
[NSNumber floatValue] is expensive, this patch removes the need for floatValue during animations by using an array of floats in an NSData container instead.


Zakay:
+1

PR #233, opened 2012-08-22
fix: Double firing of CCCallBlock in CCSequence

zeaster:
http://code.google.com/p/cocos2d-iphone/issues/detail?id=1398


Panajev:
Thanks for the patch. Did you run the actions test in Cocos2D too?

zeaster:
I have run the actions test and it looks ok to me.

PR #234, opened 2012-09-11
Added public property to allow a CCSprite subclass to forcefully disable debug drawing.

Panajev:
When using Cocos2D as an external static library, the macro used to influence certain Cocos2D features, such as Debug drawing for CCSprites, are fixed when the library is compiled.

A workaround might be to compile the library in Release and Debug versions with the DEBUG flag directing, for example, when the debug drawing macro's allow the debug shapes to be rendered.
The problem is that this might cause cases in which the debug drawing feature for CCSprite objects is not really usable: if you have a few sprites moving in front of a large background (with the background being essentially just a large CCSprite object in some instancea), all the user will see is a huge white shape and will not be able to discern the small sprites in the foreground.

If we allow the CCSprite to override the debug drawing macro at runtime, we do not either lose performance in release mode and we make debug mode more flexible preventing the case mentioned above.


Panajev:
Uhm, sorry... the other three commits are related to the support for pushing and popping scenes with transitions which I already included in a separate pull request, should I cancel that pull-request?

PR #235, opened 2012-09-14
iPhone Four Inch Display File Suffix

bentrengrove:
Added a new File Suffix to CCFileUtils to load different images when running on an iPhone with a Four Inch display.

Added test to FileUtilTests and new image of grossini to test against.


supernes:
Doh! I went ahead and did the same thing before checking if someone else has. 

My additions however provide for non-Retina widescreen images as well (there are use cases for that). Check it out -- https://github.com/supernes/cocos2d-iphone/commit/11a181aeb3234fe0b8cf050b383eb0760c5232e7

ricardoquesada:
thanks. Added iPhone5 support in this commit:
https://github.com/cocos2d/cocos2d-iphone/commit/7a953ebefbfd5eae9cb4c183decbd5e6b80c11ee

PR #240, opened 2012-10-02
Added Texture2d test showing GL state caching problem

gregharding:
This problem only happens when CC_ENABLE_GL_STATE_CACHE=1 and a single
sprite/texture is used. The asynchronous loading shouldn't affect the
test.

### DIFF
--- a/tests/Texture2dTest.h
+++ b/tests/Texture2dTest.h
@@ -199,6 +199,12 @@
 }
 @end
 
+@interface TextureAsyncBlock2 : TextureDemo
+{
+	CCSprite* sprite1_;
+}
+@end
+
 @interface TextureLibPNG : TextureDemo
 {}
 -(void) transformSprite:(CCSprite*)sprite;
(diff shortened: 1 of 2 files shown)


ricardoquesada:
@Panajev 
More info of this bug is here:
http://www.cocos2d-iphone.org/forum/topic/39169/page/3#post-188469

I also can't reproduce it, but I know the GL cache has bugs with multithreading... investigating.

ricardoquesada:
I was able to reproduce the bug by doing:
- Create an empty project
- copy & paste the code flightless posted
- disabling the FPS rendering
- enabling the GL cache.

ricardoquesada:
The bug is only present when you render just 1 sprite, so it is a minor bug... now.
But it will be a "big" bug once cocos2d supports multiple views / threads.

The bug is that from a different thread (thus different thread), the cocos2d GL cache is updated... but it shouldn't... there should be once GL cache per thread.

ricardoquesada:
I'll fix this bug in v2.2 (once multithreading / multi-view support is added )

PR #243, opened 2012-10-09
Bug with newlines in multiline labels

Zakay:
When a multiline label updates it counts multiple newlines as one and the string gets trimmed at the end.

This fixes that, keeping track of how many newlines where just skipped and appending as many characters that was skipped.

This is an extract from my code that produced the error with the last character being trimmed.
It should say "Error\n\nSome error" but it shows "Error\n\nSome erro"

```
CCLabelBMFont *message = [CCLabelBMFont labelWithString:@"Error\n\nSome error message" fntFile:@"#########.fnt"];
message.color = ccc3(0, 0, 0);
message.anchorPoint = ccp(0.5f, 1);
[message setWidth:290];
[message setAlignment:kCCTextAlignmentCenter];
message.position = ccp(dialog.contentSize.width / 2.0f, dialog.contentSize.height - 20);
[dialog addChild:message];
```

### DIFF
--- a/cocos2d/CCLabelBMFont.m
+++ b/cocos2d/CCLabelBMFont.m
@@ -552,9 +552,10 @@ void FNTConfigRemoveCache( void )
         //Go through each character and insert line breaks as necessary
         for (int j = 0; j < [children_ count]; j++) {
             CCSprite *characterSprite;
-			
-            while(!(characterSprite = (CCSprite *)[self getChildByTag:j+skip]))
-                skip++;
+            int justSkipped = 0;
+            while(!(characterSprite = (CCSprite *)[self getChildByTag:j+skip+justSkipped]))
+                justSkipped++;
+            skip += justSkipped;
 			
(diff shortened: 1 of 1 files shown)


Panajev:
Could you rewrite it  a bit not to add the forward declarations of tKerningHashElement and tFontDefHashElement to use tCCKerningHashElement and tCCFontDefHashElement please?
I think it would help it integrate better with the existing code patched here a few months ago: https://github.com/cocos2d/cocos2d-iphone/commit/6fe383a86c72ecfe6b353c59060d3a83a657c3f1

As a fellow Cocos2D user, thank you for both the patch and the code sample reproducing the issue :).

Zakay:
Sorry for that, I actually commited the wrong patch, this patch was for a mix of both master-v2 and develop-v2, I'll fix it and re-send the patch.

Panajev:
No problem :).

ricardoquesada:
Thanks. scheduled for v2.1

PR #244, opened 2012-10-10
Xcode 4.5 / iOS 6 Build Settings

colinmccune:
Minor changes to the cocos2d project build settings to provide better support with Xcode 4.5 and compiling as a static libraries. These changes already exists in the develop-v2 branch.
1. Since Xcode 4.5 / iOS6 no longer supports armv6 the cocos2d project needs to be updated so that it doesn't try to build a armv6 version.  I've remove the hard coded armv6 from the ARCHS setting and replaced it with the ARCHS_STANDARD_32_BIT value which then allows your version of Xcode to decide what ARCS should be supported. Eg: Xcode 4.5 will use armv7, armv7s and Xcode < 4.5 will use armv7 and I'm assuming even older version of Xcode will list armv6 and armv7.
2. Changed the ONLY_ACTIVE_ARCH flag from YES to NO for Debug builds. This prevents linker errors when compiling cocos2d as a static lib and doing a build against "iOS Device". Without this change only the armv7 code is compiled into the static library which will throw errors when you link it against your armv7s compiled code ("Undefined symbols for architecture armv7s"). Release builds already had this flag set to No.

### DIFF
--- a/cocos2d-ios.xcodeproj/project.pbxproj
+++ b/cocos2d-ios.xcodeproj/project.pbxproj
@@ -15212,15 +15212,10 @@
 			isa = XCBuildConfiguration;
 			buildSettings = {
 				ALWAYS_SEARCH_USER_PATHS = NO;
-				ARCHS = (
-					armv6,
-					"$(ARCHS_STANDARD_32_BIT)",
-				);
 				COPY_PHASE_STRIP = NO;
 				GCC_DYNAMIC_NO_PIC = NO;
 				GCC_OPTIMIZATION_LEVEL = 0;
 				GCC_THUMB_SUPPORT = NO;
-				ONLY_ACTIVE_ARCH = YES;
 				PRODUCT_NAME = "build all tests";
 			};
 			name = Debug;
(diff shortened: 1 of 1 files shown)


araker:
1. I don't think it's going to try to build a armv6 version, since Xcode 4.5 isn't capable of it anymore. You're right that older versions still list armv6 under valid architectures, even when armv6 isn't added to architectures. Changing it. 
2. It's set to YES to speed up debug building, I don't think it should be NO since in many cases it isn't needed. I will add a warning for this though in the release notes. 

Thanks for your commit.

colinmccune:
Thanks. A note in the release notes would be helpful.

I ran into this problem while converting our projects to Xcode 4.5 and I still don't full understand why but if I don't set ONLY_ACTIVE_ARCH = NO I get linker problems for any project that statically links cocos2d when a device is not plugged in (eg: target = "iOS Device"). When I plugin a device xcode is then able to make a better guess at choosing the active arch and our builds link correctly. This wasn't a problem in previous versions of xcode. I suspect the previous version of xcode only listed armv7 and now that armv7 and armv7s are both listed my active project and cocos2d seem to be defaulting to different active architectures.

Just a fyi, the ONLY_ACTIVE_ARCH is set to NO on the cocos2d target for the develop-v2 branch. It is still set to YES on the project level however. I can't seem to track down the commit that made that change.

Cheers,
Colin

PR #246, opened 2012-10-13
Automatically load Widescreen (4") optimized images

supernes:
Added two new suffixes -- "-wide" and "-widehd" for assets optimized for the iPhone 5 / iPod touch 5th gen "widescreen" displays.

The reason why there are two separate suffixes is that, for performance/memory usage reasons, sometimes you may not want to actually load a full-res image, even though all widescreen devices are also retina.

Images load fine, don't see the need for a test case, but I can write one if you absolutely insist =)

Also, there was a small issue with GL State caching related code, gave an error when building with state cache disabled.

### DIFF
--- a/cocos2d/Support/CCFileUtils.h
+++ b/cocos2d/Support/CCFileUtils.h
@@ -40,6 +40,8 @@
 	
 #ifdef __CC_PLATFORM_IOS	
 	NSString *iPhoneRetinaDisplaySuffix_;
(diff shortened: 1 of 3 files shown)


ricardoquesada:
Thanks. Scheduled for v2.1

ricardoquesada:
Thanks. Added iPhone5 support in this commit:
https://github.com/cocos2d/cocos2d-iphone/commit/7a953ebefbfd5eae9cb4c183decbd5e6b80c11ee

PR #247, opened 2012-10-15
Respect custom projections when converting coordinates.

slembcke:
I use custom projections in all of my universal Cocos2D projects so I can set an iPad resolution of 512x384. That along with a terrible hack so it runs with a content scale of 2x on the iPad (4x on retina iPad), it makes the iPad versions a lot easier to manage. They are pretty close to retina iPhone size, just a little bit bigger. Makes it much easier to share code and content between the two versions as the sizing all stays the same.

Anyway I won't share the terrible, fragile content scale hack, but one of the issues with using a custom projection is that the CCDirector methods don't know how to convert coordinates for them. It just does a simple flip assuming that GL coords are just flipped and calls it good. I fixed them to use the projection matrix instead.

### DIFF
--- a/cocos2d/Platforms/iOS/CCDirectorIOS.m
+++ b/cocos2d/Platforms/iOS/CCDirectorIOS.m
@@ -336,28 +336,38 @@ CGFloat	__ccContentScaleFactor = 1;
 
 -(CGPoint)convertToGL:(CGPoint)uiPoint
 {
-	CGSize s = winSizeInPoints_;
-	float newY = s.height - uiPoint.y;
-
-	return ccp( uiPoint.x, newY );
+	kmMat4 projection;
+	kmGLGetMatrix(KM_GL_PROJECTION, &projection);
+	
+	kmMat4 inv_projection;
+	kmMat4Inverse(&inv_projection, &projection);
+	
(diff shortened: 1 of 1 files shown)


ricardoquesada:
I tried the patch but it breaks the conversions on cocos2d 2d projection.
If you try any test (eg:"menu test") the touches will have the wrong coordinates

Any idea?

slembcke:
Oh. Hrm. Really? I thought I tested it on regular projections earlier...

Sorry about that, I'll take another look at it.

ricardoquesada:
Ok. Thanks. If could fix this issue, I will happily apply the pull request.

slembcke:
Yup. I haven't forgotten, I just haven't had time to revisit it yet. Probably tomorrow though.

slembcke:
Oh, shoot. I see what the issue is. The default projection is a 3D one that also contains a camera transformation in the modelview matrix. It's probably _fairly_ low risk to read the modelview matrix back as the camera matrix wherever the conversion methods are called. If somebody was doing something weird and calling them from a visit or draw method it wouldn't work though.

Let me think about it a little more. I'll get back to you.

ricardoquesada:
Sure, let's try that approach.

PR #249, opened 2012-10-17
CCAnimate action improvements

rsanchezsaez:
##### Description

Added new designated initialiser

```
+(id)actionWithAnimation:(CCAnimation*)action startFrameIndex:(NSInteger)start
```

It allows choosing an animation start frame (and will cycle until it reaches the chosen frame minus one).

Added

```
+(id)actionWithRandomStartAnimation:(CCAnimation*)action
+(id)actionWithRandomStartAnimation:(CCAnimation*)action factor:(float)factor
```

They allow you to randomise the animation cycle. Useful for starting several animation actions at the same time but having them play unsynchronized.

Added Test for the new methods to the SpriteTest suite (SpriteAnimationSplitRandom). There's a delay between subsequent animation loops. That's intended, in order to notice the randomized start frame in each iteration.

Also made a small fix/optimization to CCAnimate update (the 'break' was not in the correct place, so it would not skip frames correctly on low framerates).
##### Examples

Given an 'animation' with 5 frames, running

```
[CCAnimate actionWithAnimation:aniation startFrameIndex:3]
```

will show frames in this order: 3, 4, 0, 1, 2

Given an 'animation' with 10 frames, running

```
[CCAnimate actionWithRandomStartAnimation:animation factor:0.2]
```

 will chose a random startFrameIndex within the range [0, 1].

Given an 'animation' with 10 frames, running

```
[CCAnimate actionWithRandomStartAnimation:animation factor:0.4]
```

will chose a random  startFrameIndex within the range [0, 1, 3, 4].


rsanchezsaez:
I added a CCScheduler fix to this pull request.

CScheduler: Fixed issue that prevented to a method to schedule itself from within a scheduled callback. See http://www.cocos2d-iphone.org/forum/topic/23882 for discussion.

ricardoquesada:
Thanks. Nice addition. Scheduled for v2.1

rsanchezsaez:
Thanks Ricardo. It is great to be able to contribute to cocos2d!

Panajev:
Awesome addition!

[web [iPhone msg]]

PR #253, opened 2012-10-23
Fixed CCLabelBMFont contentSize problem

mikedaley:
When using bitmap fonts that have large strokes or shadows, the contentSize calculated for the CCLabelBMFont instance was too narrow and the string would leak our of the right hand side of the contentSize bounding box. This fix addresses that issue causing the contentSize to be correctly calculated.

Also added a new test in LabelTest.m called LabelBMFontBounds


ricardoquesada:
thanks. merged in develop-v2 branch.
https://github.com/cocos2d/cocos2d-iphone/commit/3e44487a8b93fe8d6664592a786c97966a0f9a4d

PR #254, opened 2012-10-24
CCLabelBMFont bounding box fix

mikedaley:
Submitted in error


ricardoquesada:
this was applied in `develop-v2` branch. thanks.

PR #256, opened 2012-10-25
Objective chipmunk integration

slembcke:
Including chipmunk.h in the headers when Chipmunk integration was turned on was causing problems for including Objective-Chipmunk in a project. It would try to include chipmunk.h a second time and cause issues because of the CP_ALLOW_PRIVATE_ACCESS stuff.

I changed it to use forward declarations of the Chipmunk structs instead.

### DIFF
--- a/cocos2d/CCPhysicsDebugNode.h
+++ b/cocos2d/CCPhysicsDebugNode.h
@@ -22,17 +22,10 @@
 #import "ccConfig.h"
 
 #if CC_ENABLE_CHIPMUNK_INTEGRATION
-
-#ifdef CP_ALLOW_PRIVATE_ACCESS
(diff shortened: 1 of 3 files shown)


ricardoquesada:
Is there another way to fix this ?
I still need to use, for the JS bindings,  `cpSpace*` instead of `struct cpSpace*` in the declaration of the method.

The problem is that BridgeSupport generates a different signature when `struct cpSpace*` is used,  and the JS bindings break.

slembcke:
Hrm. So you can do a forward declaration of a struct but you can't have duplicate typedefs. Let me tinker with it some more and I'll make a new pull request.

ricardoquesada:
ok. thanks.

ricardoquesada:
Closing this one since it was superseded by #257

PR #257, opened 2012-10-26
Alternate fix for the Objective-Chipmunk integration issues.

slembcke:
Ok, so I found an alternate solution that should work for both of us then. I disabled private access in CCPhysicsDebugNode.h and instead used the CP_PRIVATE() macro in the .m file. I also had to make some small changes to Objective-Chipmunk.h to better deal with including chipmunk.h twice.

I also fixed a small bug where shapes attached to rogue bodies would cause the debug layer to crash.


pkclsoft:
That correction to the debug layer was especially useful.  Thanks

PR #259, opened 2012-10-30
use orphaning + glMapBuffer in the ParticleSystem to avoid stalling the ...

kirillkh:
...pipeline while waiting for the GPU to sync the VBO

### DIFF
--- a/cocos2d/CCParticleSystemQuad.m
+++ b/cocos2d/CCParticleSystemQuad.m
@@ -445,7 +445,10 @@
 {
 #if CC_USES_VBO
 	glBindBuffer(GL_ARRAY_BUFFER, quadsID_);
-	glBufferSubData(GL_ARRAY_BUFFER, 0, sizeof(quads_[0])*particleCount, quads_);
+	glBufferData(GL_ARRAY_BUFFER, sizeof(quads_[0])*totalParticles, nil, GL_STREAM_DRAW);
+	void *buf = glMapBufferOES(GL_ARRAY_BUFFER, GL_WRITE_ONLY_OES);
+	memcpy(buf, quads_, sizeof(quads_[0])*particleCount);
+	glUnmapBufferOES(GL_ARRAY_BUFFER);
 	glBindBuffer(GL_ARRAY_BUFFER, 0);
 #endif
 }


Panajev:
Very interesting, have you managed to benchmark it?

[web [iPhone msg]]

kirillkh:
I haven't benchmarked the particles specifically, but in my case with 22 emitters the overall benefit is fabulous: 45% off the time per frame (22ms -> 12ms), and I have a fair amount of other stuff happening aside from particles. And I'm not sure a synthetic benchmark would be very illuminating in any case, because the command pipeline stalling can have varying effect depending on the workload.

Panajev:
You are correct, also since it seems to be giving you quite a fantastic
performance increase, unless the regular Cocos2D provided particles test
(just add tons of particles and see what happens to the FPS) show some
nasty corner cases, this patch seems quite good as reducing command buffer
stalls is a very good thing to do ;).

kirillkh:
Can you point to a specific test in the suite? Also note that I don't have the resources to test it properly across devices, as I only have access to ipad 1 and ipad 3 here with just a handful of iOS releases. So I think the maintainers should benchmark it properly across devices and iOS versions.

P.S. glMapBuffer is not required to get rid of the stalls - you can still use glBufferSubData, so in case you run into a nasty corner case you can try replacing glMapBuffer with glBufferSubData. However, my limited testing shows that glMapBuffer does give a performance benefit of its own, especially in conjunction with orphaning. I don't have an explanation for it, but I've used this approach before, and in both cases glMapBuffer gave an improvement.

Panajev:
iPad 1 and iPad 3 are the best devices to test this one. The former was
quite heavily pixel processing limited and the latter quadruples the
resolution (with a similar increase in efficiency and computing resources,
bit not the same).

those two are the devices I would expect performance issues to crop up
actually.

The test is called "Particle Test" any of them should do.

araker:
I've tested glMapBuffer a while ago and it performed worse on armv6 devices. But that was on 4.2 I believe, so I can retest this to see if things have changed. The call to glBufferSubData should be changed anyway, since that call blocks (but has benefits if you only update parts of the buffer, which isn't the case here), while glBufferData just overwrites the buffer.

kirillkh:
I didn't see any regressions in the tests, but also didn't see much of an improvement either. Apparently, stalls have little influence in this workload. I did manage to modify one of the tests in a way that showed some benefit (33 fps -> 40 fps) with orphaning by modifying the emitter parameters to match what we use and instantiating many emitters simultaneously.

Apparently, our use case is with much simpler individual particles that require little CPU or GPU time, but we have relatively many of them.

kirillkh:
I have also noticed that I can't use GL_STREAM_DRAW in the test environment for some reason - the OpenGL ES Analysis tool says "Invalid enum for argument 'usage'". Probably I'm not configuring something correctly, but I have no idea what.

ricardoquesada:
Interesting. I will apply it to develop-v2 too.
Yes, I was told that `glBufferSubData` could have performance issues.

araker:
GL_STREAM_DRAW is not available in OpenGL ES 1.1.. http://www.khronos.org/opengles/sdk/1.1/docs/man/

kirillkh:
Oh. Wrong branch, then? The patch was intended for cc2.0, as that's what I'm working with. Sorry for this confusion, I'm working with a local copy of cc2d 2.0 and (after looking at the FAQ) thought "develop" was the branch with the most recent code for it.

araker:
yeah for v2 it's develop-v2.

ricardoquesada:
Thanks. Patch applied in v2.
https://github.com/cocos2d/cocos2d-iphone/commit/edf91b453d1fccb972ede6cb9bece0a649cf415a

PR #262, opened 2012-11-06
Add support for handling OSX gesture event

pierredavidbelanger:
A little addition to be able to use OSX gesture.

It will compile and run on OSX 10.5.2 and later.

At runtime It should not create any problem with installation that do not supports gesture since the `NSResponder`'s gesture selectors will simply not be called.

In this commit:
- `CCGLView` delegates gesture events to `CCEventDispatcher`
- new `CCGestureEventDelegate` protocol to support: beginGesture, magnify, smartMagnify, rotate, swipe and endGesture, it supports swallowing event too
- add/remove/removeAllGestureDelegate selectors in `CCEventDispatcher` with priority
- gestureEnabled and gesturePriority properties in `CCLayer`
- edited `TouchTest` of the EventsTest target to demonstrate gestures usage

I tried to respect the cocos2d naming convention, but let me know if your are not ok with the way I named the new public protocol and selectors, I will refactor to suit your taste.

### DIFF
--- a/cocos2d/CCLayer.h
+++ b/cocos2d/CCLayer.h
@@ -93,7 +93,7 @@ typedef enum {
 #elif defined(__CC_PLATFORM_MAC)
 
 
-@interface CCLayer : CCNode <CCKeyboardEventDelegate, CCMouseEventDelegate, CCTouchEventDelegate>
+@interface CCLayer : CCNode <CCKeyboardEventDelegate, CCMouseEventDelegate, CCTouchEventDelegate, CCGestureEventDelegate>
 {
 	BOOL		mouseEnabled_;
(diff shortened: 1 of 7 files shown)


pierredavidbelanger:
A little note here about the swipe gesture.

The `NSResponder swipeWithEvent:` selector will be called only if, in the OSX Trackpad System Preferences:
- Swipe is enabled and set to "Swipe with 3 fingers" or "Swipe with 2 or 3 fingers"
- and "Drag with 3 fingers" is disabled

I guess this is because OSX need to be sure its a swipe and not a quick scroll or a drag before sending the swipe gesture to the responder.

PR #263, opened 2012-11-06
Fix some small problems for CCScheduler

HarrisonXi:
1. Remove dup initialization for repeat
2. Check paused state for target without custom selectors correctly
3. Avoid adding one target for twice into idsWithSelectors (in function
   pauseAllTargets).


ricardoquesada:
Thanks. Do you have a test case for the "avoid adding the target twice" ? thanks.

ricardoquesada:
Ah... and please, send the pull request against "develop-v2", and not "master-v2", since a lot of things changed since then. Thanks.

HarrisonXi:
Sorry, I forgot to change the branch. I will add test case and pull it to develop-v2. Thanks for your reminding.

PR #264, opened 2012-11-06
Fix some small problems for CCScheduler

HarrisonXi:
1. Add runForever check before cancel timer automatically.
2. Check paused state for target without custom selectors correctly.
3. NSSet has avoided adding dup item already. Added a simple test for
   this.


ricardoquesada:
Thanks. The patch seems to be correct.
But I couldn't trigger the bug that the patch fixes with the new test.

PR #267, opened 2012-11-09
Cleanup scheduled actions of a CCTransitionRadialCCW that is popped

vargusz:
Cleanup scheduled actions of a CCTransitionRadialCCW that is popped before finishing its animation.

When a CCTransitionRadial\* transition scene is popped from the director before the animation completes, the layerAction's CCCallFunc action will never be called, but it retains a reference to its target object, so neither the transition scene, nor its outgoing and incoming scenes will ever be deallocated.

By cleaning up the scheduled actions and animations of the CCProgressTimer instance (identified by the kSceneRadial tag), we can get rid of this reference.

### DIFF
--- a/cocos2d/CCTransitionRadial.m
+++ b/cocos2d/CCTransitionRadial.m
@@ -98,7 +98,7 @@ enum {
 -(void) onExit
 {
 	// remove our layer and release all containing objects
-	[self removeChildByTag:kSceneRadial cleanup:NO];
+	[self removeChildByTag:kSceneRadial cleanup:YES];
 	[super onExit];
 }
 @end


ricardoquesada:
Thanks. Patch applied in v2.1
https://github.com/cocos2d/cocos2d-iphone/commit/b794d6686c4467544cf8d1b067b0a8734249bde5

vargusz:
Thanks for adding me to the AUTHORS file! However, could you please change my name to "Gábor Varga"? Thank you.

ricardoquesada:
@vargusz 
sure. done.

ricardoquesada:
Reopening it since this patch was for v1.1

PR #268, opened 2012-11-12
Proper disabling of events during transitions

vargusz:
Touch events were disabled when transition objects were initialized, and reenabled when they successfully finished. Initialization may be done in advance, so it's too early to disable events here. Successful finishing is not guaranteed, as transition scenes may be popped from the director.

The correct place for these disable/enable events would be the onEnter and onExit methods of the CCTransition class.


ricardoquesada:
thanks. Patch applied in v2.1
https://github.com/cocos2d/cocos2d-iphone/commit/ee7133fe281413a7f832c2ca1b405bf33d547c2b

Leaving PR open for v1.1

PR #270, opened 2012-11-14
CCRotateTo action could cause errors when rotation

LamPham:
Fixed an issue where the x and y rotational values were not the same when we use the CCRotateTo action.

Now we make sure x and y never diverges if we want it to rotate with the same angle.

Mentioned in:
http://www.cocos2d-iphone.org/forum/topic/56028?replies=3#post-220127


ricardoquesada:
thanks. scheduled for v2.1

PR #271, opened 2012-11-15
Develop v2

HarrisonXi:
Fix firstGID parser issue for new version TMX file with external TSX file

In latest Tiled, there is no firstGID property in external TSX file, so we need to get the correct firstGID in TMX file.
Had fixed this issue and tested it in TiledMapTest.TMXOrthoTest2.


ricardoquesada:
Thanks. Scheduled for v2.1

ricardoquesada:
thanks. fixed in:
https://github.com/cocos2d/cocos2d-iphone/commit/c03af5c7969c8acfdf9cc19354333d659e503b4c

PR #275, opened 2012-12-17
Stackable CCMoveBy and CCMoveTo actions

rsanchezsaez:
I have rewritten the CCMoveBy and CCMoveTo actions so they are stackable.
### Description

This means that you can now run two simultaneous CCMoveBy/CCMoveTo actions on the same CCNode, and the node will perform a movement equal to the result of applying both movements at the same time.
### Uses

This allows for several cool effects. Some examples:
- Have a looping CCMoveBy action constantly running inside a CCRepeatForever action, and still be able to move the node around the screen with additional CCMoveBy actions (think, for example, of a balloon that has a looping floating animation and also moves around the screen).
- Having a character being affected by different move forces (think of a character affected by the wind and by the player controlled movement at the same time).
### Performance

I added CCMoveBy/CCMoveTo tests to the PerformanceTest-Sprite target. I ran tests comparing the new implementation to the previous one. The new implementation doesn't affect performance whatsoever.

You can compare performance of both sets of actions by checking out commit 3eceb6831f6151cfda1e2beb618e47fa3284145e (in this commit, my implementations are called CCMoveByEx and CCMoveToEx).
#### Details

In order to make movement actions stackable, CCMoveBy must be the base action. The new implementation makes CCMoveBy the base action, and CCMoveTo a child of CCMoveBy. CCMoveTo just calculates a CCMoveBy based on the current node position. The old implementation was the other way around. 

The new CCMoveBy implementation uses a new CCNode helper method called

```
- (void)moveBy:(CGPoint)positionDelta {
    position_ = ccpAdd(position_, positionDelta);
    isTransformDirty_ = isInverseDirty_ = YES;
}
```

which also allows for writing cleaner code regarding positioning of CCNodes that usually do relative (as opposed to absolute) movements.
#### Previous behaviour

The previous behaviour when executing two CCMoveBy actions at the same time was that only the latest executed action affected the sprite. This resulted in a somewhat awkward behaviour when, for example, there was a very long CCMoveBy action being run and you executed a new shorter CCMoveBy action without stopping the previous one first. In this case, the longer action would stop being applied while the new shorter one was running, but then it would weirdly resume after the new one finished.

### DIFF
--- a/cocos2d-ios.xcodeproj/project.pbxproj
+++ b/cocos2d-ios.xcodeproj/project.pbxproj
@@ -5284,7 +5284,7 @@
 		508CA7E711932A1F003AC397 /* SchedulerTest.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; name = SchedulerTest.h; path = tests/SchedulerTest.h; sourceTree = "<group>"; };
 		508EAC2F1193F8B0007F058D /* utlist.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = utlist.h; sourceTree = "<group>"; };
 		509807081175041800EA7266 /* libbuild all tests.a */ = {isa = PBXFileReference; explicitFileType = archive.ar; includeInIndex = 0; path = "libbuild all tests.a"; sourceTree = BUILT_PRODUCTS_DIR; };
-		509A79950F6188420032F449 /* CCSprite.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; lineEnding = 0; path = CCSprite.h; sourceTree = "<group>"; xcLanguageSpecificationIdentifier = xcode.lang.objcpp; };
+		509A79950F6188420032F449 /* CCSprite.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; lineEnding = 0; path = CCSprite.h; sourceTree = "<group>"; };
 		509A79960F6188420032F449 /* CCSprite.m */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.objc; lineEnding = 0; path = CCSprite.m; sourceTree = "<group>"; };
 		509A7A970F61A2400032F449 /* grossini_dance_atlas.png */ = {isa = PBXFileReference; lastKnownFileType = image.png; path = grossini_dance_atlas.png; sourceTree = "<group>"; };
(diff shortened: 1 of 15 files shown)


ricardoquesada:
Looks good. Do you have a test case for it (besides the Performance tests) ? Thanks.

rsanchezsaez:
@ricardoquesada I have added a test case to the SpriteTest suite.

It shows two sprites on screen and runs these sets of actions:

##### Sprite A

```
[sprite runAction:
 [CCRepeatForever actionWithAction:
  [CCSequence actions:
   [CCMoveBy actionWithDuration:6 position:ccp(200, 0)],
   [CCMoveBy actionWithDuration:6 position:ccp(-200, 0)],
   nil]]];

[sprite runAction:
 [CCRepeatForever actionWithAction:
  [CCSequence actions:
   [CCDelayTime actionWithDuration:1],
   [CCMoveBy actionWithDuration:2 position:ccp(0, 100)],
   [CCDelayTime actionWithDuration:1],
   [CCMoveBy actionWithDuration:2 position:ccp(0, -100)],
   nil]]];
```

##### Sprite B

```
[sprite runAction:
 [CCRepeatForever actionWithAction:
  [CCSequence actions:
   [CCMoveTo actionWithDuration:6 position:ccpSub(p, ccp(200, 0))],
   [CCMoveTo actionWithDuration:6 position:p],
   nil]]];

[sprite runAction:
 [CCRepeatForever actionWithAction:
  [CCSequence actions:
   [CCDelayTime actionWithDuration:1],
   [CCMoveBy actionWithDuration:2 position:ccp(0, -100)],
   [CCDelayTime actionWithDuration:1],
   [CCMoveBy actionWithDuration:2 position:ccp(0, 100)],
   nil]]];
```

ricardoquesada:
@rsanchezsaez 
Thanks!

ricardoquesada:
Thanks. merged.

I also added stackable support for `CCJumpBy` and `CCJumpTo` and changed a bit the way to compute the new position.
Instead of using `_previousTick` I am using `_previousPosition`.
`_previousTick` is faster but could lead some some inaccuracies due to the floating point issues.

rsanchezsaez:
Thanks for the merge.

The changes you made to make the rest of move actions stackable are very nice. I also understand the rationale for using _previousPosition, cool stuff.

Two additional notes:

a) I forgot to remove some
   @ since v2.1beta2-custom
tags that I added to the CCMoveBy and CCMoveTo header comments. I think these should be removed (or replaced by something along the lines of: "Move actions are stackable since v2.1".

b) grossini_dance_atlas.png and grossini_dance_atlas-hd.png should be added to the ActionTest target, otherwise the stackable tests will crash.

Cheers!
Ricardo

ricardoquesada:
Thanks.

PR #277, opened 2013-01-02
Protect CCConfiguration accessor with sync lock

tchaikov:
i ran into a race condition where two invocations of [CCConfiguration init] were running in two threads. because i was calling [CCGLView viewWithFrame: ...] and using CCFileUtils in a background thread. the later tries to use [CCConfiguration sharedConfiguration]. but CCConfiguration's singleton initializer is not protected by a lock, so there is chance that both thread try to create an instance of it.


tchaikov:
meant to be merged into develop-v2

PR #283, opened 2013-01-22
Customizable touchArea for CCMenuItem

sdabet:
- Added 'touchArea' property to CCMenuItem, which is initialized to the item's outside box in 'setContentSize'
- Use 'touchArea' in 'itemForMouseEvent' and 'itemForTouch' in CCMenu (instead of 'rect')
- Removed 'rect' from CCMenuItem
- Added corresponding tests


ricardoquesada:
Thanks. What do you think of renaming `touchArea` to `areaSize` ? Since `touch` is OK, it might not make sense on OS X.
Or if you have a better name, please,  let me know.

sdabet:
Yep, `touch` doesn't really make sense for OS X. But `areaSize` would suggest it is a `CGSize` whereas it's a `CGRect`... 

How about `activeArea` ?

ricardoquesada:
+1

PR #285, opened 2013-01-25
add loop sound effect support for simple audio engine

michaelochs:
This enables the simple audio engine to play sound effects in a loop until you call stopEffect: with the returned sound id.


sdabet:
Thanks for sharing, I'm actually surprised it was not available yet

njt1982:
Me too - why is this PR not merged? Seems strange to not include looping on effects (but looks like background music can loop)....

ghost:
This won't be merged because it has a bug.  Try playing 32 sounds after looping your sound effect - the source for the looped sound will be recycled which effectively kills your looped sound.  CDSoundSource was created specifically for looping sounds with SimpleAudioEngine as it locks the underlying source.

PR #286, opened 2013-02-05
Add generic scene method to CCLayer

xissburg:
Added a generic scene method to CCLayer.

### DIFF
--- a/cocos2d/CCLayer.h
+++ b/cocos2d/CCLayer.h
@@ -138,6 +138,9 @@ typedef enum {
 
 #endif // mac
 
+/** Returns a scene containing an instance of this layer as its only child. */
++ (CCScene *)scene;
+
 @end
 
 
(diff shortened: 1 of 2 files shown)


ricardoquesada:
Thanks. But I prefer to add to the API only the "must" have methods.
Although `+scene` could be handy under certain circumstances, I prefer not to add it.
Thanks again.

rakkarage:
how about something like this:

```
@interface CCNode (extensions)
+ (id)nodeWithScene;
@end

@implementation CCNode (extensions)
+ (id)nodeWithScene
{
    CCScene *scene = [CCScene node];
    [scene addChild:[self node]];
    return scene;
}
@end
```

so everyone does not have to implement + (id)scene in each class

ricardoquesada:
Sorry, but I am not convinced.

PR #252, opened 2012-10-22
Clipping node

pierredavidbelanger:
Here is a new type of node: CCClippingNode.

This node can be used to restrict the drawing of some content (its childs) inside a stencil (an other node and its childs), also the clipping node can be used in inverted mode to restrict the drawing to the outside of the stencil instead of inside.

This node uses 
- the graphic card stencil buffer to achieve fast clipping.,
- and can optionally treat pixel bellow an alpha threshold as an absence of pixel, using the graphic card alpha test on OSX or shaders on iOS, thus allowing one to use as a stencil an image with alpha.

This branch contains:
- the code to enable the stencil buffer on iOS OES framebuffer
- a new node (CCClippingNode)
- a set of tests and demos + a new target for those tests

Everything works on iOS and OS X, using optimizations on OS X where possible.

Do not hesitate to let me know what you think.

I will stop working on this branch for now unless you need me to commit something else.

Thank you, and I hope you and the community will find this little addition as useful as I am.

PR #290, opened 2013-02-21
Use glClear to clear the stencil buffer.

colesbury:
Use glClear instead of drawing a rectange so that current layer's mask
for the entire stencil buffer is cleared. This fixes clipping when
drawn to a texture larger than the screen size or when the stencil or
clipping node's children contain vertices with negative coordinates.

Fixes issue 1463  http://code.google.com/p/cocos2d-iphone/issues/detail?id=1463
(CCClippingNode does not clear the stencil buffer properly)

I've checked the clipping tests and they look ok.

### DIFF
--- a/cocos2d/CCClippingNode.m
+++ b/cocos2d/CCClippingNode.m
@@ -215,17 +215,15 @@ static void setProgram(CCNode *n, CCGLProgram *p) {
     ///////////////////////////////////
     // CLEAR STENCIL BUFFER
     
-    // manually clear the stencil buffer by drawing a fullscreen rectangle on it
     // setup the stencil test func like this:
-    // for each pixel in the fullscreen rectangle
+    // for each pixel in the stencil buffer
     //     never draw it into the frame buffer
     //     if not in inverted mode: set the current layer value to 0 in the stencil buffer
     //     if in inverted mode: set the current layer value to 1 in the stencil buffer
     glStencilFunc(GL_NEVER, mask_layer, mask_layer);
+    glClearStencil(!_inverted ? 0 : ~0);
     glStencilOp(!_inverted ? GL_ZERO : GL_REPLACE, GL_KEEP, GL_KEEP);
-    
-    // draw a fullscreen solid rectangle to clear the stencil buffer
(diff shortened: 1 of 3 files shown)


pierredavidbelanger:
I understand that http://code.google.com/p/cocos2d-iphone/issues/detail?id=1463 need to be fix, but I think it will not be a good idea to revert using glClear here.
This is what I was doing in the first implementation of CCClippingNode, and it did not work on some recent MBP hardware.

Here is the commit where I swapped the glClear method for a fullscreen rectangle:
https://github.com/cocos2d/cocos2d-iphone/commit/1611da9702d030f5a9767fd7a02ce8ace959b0bd#cocos2d/CCClippingNode.m

colesbury:
Thanks for your quick reply!  I'd love to get this fixed.

Do you remember which tests didn't work on MBP when using glClear?

pierredavidbelanger:
Here is what the "Raw Stencil Tests" looks like on OS X (OS X 10.7.4, AMD Radeon HD 6750M) with the glClear:

https://github.com/cocos2d/cocos2d-iphone/pull/252#issuecomment-9821861

I, my self, was not able to reproduce this problem. @ricardoquesada helped me debug and fix the issue.

colesbury:
Sounds like it's important to work around that AMD driver bug.  I've updated the pull request to use ccDrawRect on OS X and added a test case that demonstrates the original problem I was having.

ricardoquesada:
Thanks @colesbury 

@pierredavidbelanger If you agree with the patch, I will merge it.

pierredavidbelanger:
I tested the patch in the simulator, an iPhone 4, an iPad 2 and a mid-2009 MBP, everything looks good.
Except that I can not confirm that the bug with glClear on some ATI drivers is still inexistant; but by reading the code, we can tell it should still be ok.

I think it is a good idea to revert using glClear on iOS.

Good job @colesbury

pierredavidbelanger:
Oups, obviously that the bug with glClear on OSX is still inexistant: glClear is called only when on iOS ;)

ricardoquesada:
Thanks .I will merge it.

PR #292, opened 2013-02-24
Enable full colors in simulator

iMobCoding:
Other templates should be changed too


ricardoquesada:
Thanks. Why would you want different color buffers in the simulator and the device ?

iMobCoding:
Because simulator and real device obviously differ in how they represent 565 RGB model. I am talking about color banding in simulator if kEAGLColorFormatRGB565 is used. Please check this topic http://www.cocos2d-iphone.org/forum/topic/234276

ricardoquesada:
Thanks!
I will do a few more tests, and if it behaves as said on the forum thread, I will merge it.

ricardoquesada:
Finally what I did was to improve the description on the templates
https://github.com/cocos2d/cocos2d-iphone/commit/0c14a38d9d17bea565b5ccfdad89a5744ed81b57

PR #293, opened 2013-02-26
Fix for using Objective-Chipmunk with CC_PHYSICS_INTEGRATION.

slembcke:
I've been telling people to work around this issue for a while, but I came up with a much simpler fix that shouldn't interfere with the JS binding generator.


slembcke:
Awesome. Thanks. That should make a few peoples day.

ricardoquesada:
I updated the commit:
https://github.com/cocos2d/cocos2d-iphone/commit/73ba1cde2d9f4bed869f713171579a75979473a4

If you notice something wrong, please let me know. thanks.

PR #295, opened 2013-03-01
Added initWithFile method to CCParticleSystemQuad for sprite sheets

NachtRaveVL:
Added initWithFile method to CCParticleSystemQuad so that a sprite sheet may be used directly from a dictionary without extra code. This allows one to simply use a CCSpriteFrameCache named object in place of the textureFileName so that the particle texture is used from a sprite sheet, thus increasing efficiency of CCParticleBatchNode and CCParticleSystemQuad when different sprites are to be used but are all on the same sprite sheet texture.

### DIFF
--- a/cocos2d/CCParticleSystemQuad.m
+++ b/cocos2d/CCParticleSystemQuad.m
@@ -38,6 +38,7 @@
 #import "CCTextureCache.h"
 #import "ccMacros.h"
 #import "CCSpriteFrame.h"
+#import "CCSpriteFrameCache.h"
 #import "CCDirector.h"
 #import "CCShaderCache.h"
 #import "ccGLStateCache.h"
@@ -49,6 +50,7 @@
 #import "Support/CGPointExtension.h"
 #import "Support/TransformUtils.h"
 #import "Support/NSThread+performBlock.h"
+#import "Support/CCFileUtils.h"
 
 // extern
 #import "kazmath/GL/matrix.h"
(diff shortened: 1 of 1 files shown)


ricardoquesada:
Thanks.
It is a nice feature. Will it work if the SpriteFrame is rotated, or if its has been trimmed ? 

I think this method will be easier to understand:
`initWithSpriteFrame:(CCSpriteFrame*)frame`

And if the frame is rotated or trimmed, abort with an assert.

What do you think?

NachtRaveVL:
I haven't really an indepth knowledge of the dynamics of CCSpriteFrame, but I am assuming since it's just doing a rect set that the rect that is returned from CCSpriteFrame needs to be pretty much ready to go for texturing coordinates (i.e. not rotated, already accounting for trim, etc.).

The reason I went with initWithFile:/initWithDictionary: is because we're dropping a lot of plists of various effects into the game, and it's just a lot easier to have their textureFileName field pointing over to a CCSpriteFrame in the CCSpriteFrameCache. This makes it so that instead of coding a few multiple lines every instantiation we just instantiate it with the plist there and then in one line of code and it automagically hooks up to CCSpriteFrameCache, e.g.:

```
        CCParticleSystemQuad *coinShock1 = [[CCParticleSystemQuad alloc] initWithFile:@"coin_ps_shockwave1.plist"]; // uses coin_ring1.png, a part of the coin_particles.png sprite sheet
        CCParticleSystemQuad *coinShock2 = [[CCParticleSystemQuad alloc] initWithFile:@"coin_ps_shockwave2.plist"]; // uses coin_ring2.png, a part of the coin_particles.png sprite sheet
        CCParticleSystemQuad *coinBurst1 = [[CCParticleSystemQuad alloc] initWithFile:@"coin_ps_burst1.plist"]; // uses coin_sparkle1.png, a part of the coin_particles.png sprite sheet
        CCParticleSystemQuad *coinBurst2 = [[CCParticleSystemQuad alloc] initWithFile:@"coin_ps_burst2.plist"]; // uses coin_sparkle2.png, a part of the coin_particles.png sprite sheet

        // batch nodes with coin_particles.png sprite sheet, renders all particle systems, which use different sprites but all in the same sprite sheet, in one call
        [_coinPlayParticleSystemBatchNode addChild:coinShock1];
        [_coinPlayParticleSystemBatchNode addChild:coinShock2];
        [_coinPlayParticleSystemBatchNode addChild:coinBurst1];
        [_coinPlayParticleSystemBatchNode addChild:coinBurst2];
```

ricardoquesada:
By using `CCSpriteFrame` objects, or better `CCSpriteFrameName` objects, you could do the same:

``` obj-c
// make sure that the sprite frames are not rotated, and not trimmed
[[CCSpriteFrameCache shared....] addSpriteFramesFromFile:@"coins.plist"];

CCParticleSystemQuad *coinShock1 = [[CCParticleSystemQuad alloc] initWithSpriteFrameName:@"coin_ring1.png.plist"]; 
CCParticleSystemQuad *coinShock2 = [[CCParticleSystemQuad alloc] initWithSpriteFrameName:@"coin_ring2.png.plist"]; 
```

What do you think about it ?

NachtRaveVL:
Perhaps initAsSpriteFrameWithFile:/initAsSpriteFrameWithDictionary: ???

The trouble that was mainly occurring was that the particle system is initialized via plist, of which the initWithDictionary: code is only inside of CCParticleSystem. In that method it is using textureFileName as a pure link to a CCTexture2D object (from CCTextureCache), hence the need for the redirectDict in the original code (so it points against the sprite sheet's CCTexture2D object, but can still init the particle system with the plist).

Aside from modifying that code (which would also work technically), this seemed just a lot quicker.

PR #303, opened 2013-04-02
Updating Image to be Less Offensive

burtlo:
While working with the various tests tiled map formas and images I came across the Isometric map which had a over-sexualized female character in the image tileset.

This minor update removes the unnecessary detail by the artist on the character's chest.


mikelovesrobots:
+1

ricardoquesada:
This is the 1st time that I notice it.

PR #306, opened 2013-04-12
Fixed issue with text alignment on multi-line auto-sized CCLabelTTF's

richard-sim:
Added new init method to CCTexture2D that doesn't take dimensions, so that CCLabelTTF can create auto-sized labels that specify alignment (otherwise multi-line labels will always have the default alignment, which isn't desired). CocosBuilder already behaves in this way, so this makes the two of them match, which they did not previously.


richard-sim:
Erk, included module-related changes. Closing & will create a new pull request without those changes.

richard-sim:
Reopened as there doesn't appear to be a simple way to make a pull request for individual changes. boo!

.gitmodules
```diff
@@ -15,4 +15,4 @@
 	url = git://github.com/cocos2d/cocos2d-js-tests.git
 [submodule "external/CCBReader"]
 	path = external/CCBReader
-	url = git://github.com/cocos2d/CCBReader.git
```
richard-sim, line 18: Ignore this file...

external/CCBReader
```diff
@@ -1 +1 @@
-Subproject commit 2f9d46dc959d0ec2c8dcbdc37243df39c0c1c454
```
richard-sim, line 1: Ignore this file too...

PR #308, opened 2013-05-06
Implemented CCEasePolynomial Action.

nader-eloshaiker:
- Developed a replacement for exponential mathematical routine with a nth order polynomial which resembles an exponential curve for values of 0 <= t <= 1 and an order of 6.
- Advantage of using polynomial curves is that they do not asymptote.
- Developers can control the acceleration at which it eases in or out by adjusting the polynomial value, polynomial must be greater than 1.
- Allowed for odd and even values of polynomial orders as odd values will exhibit an inflection in the curve.

The attached image shows the timing curves for the following:
- Purple = Ease In Polynomial order 3
- Green = Ease In Polynomial order 9
- Blue   = Ease In Polynomial order 6
- Red    = Ease In Exponential

ccactionease

### DIFF
--- a/cocos2d/CCActionEase.h
+++ b/cocos2d/CCActionEase.h
@@ -2,6 +2,7 @@
  * cocos2d for iPhone: http://www.cocos2d-iphone.org
  *
  * Copyright (c) 2008-2009 Jason Booth
+ * Copyright (c) 2013 Nader Eloshaiker
  *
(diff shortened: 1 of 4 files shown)


nader-eloshaiker:
I noticed that there aren't many names/authors in the comment header of code (CCEaseAction.m). Am I putting my name in the wrong place?

ricardoquesada:
Your name is good. thanks.

PR #321, opened 2013-08-06
Add support to ccGLStateCache for glBlendFuncSeparate

combinatorial:


slembcke:
Sorry for the blast from the past. The v3.1 renderer will support separate blending modes and the ccGL\* functions have gone away.

PR #340, opened 2013-09-22
Develop v3 panajev

Panajev:
Cocos2D now also produces a fat static library that can easily be distributed and embedded for both iOS and OS X projects (you can also more easily keep track of the version of the distributed binary) through new targets in both projects. This has necessitated a reorganization of the project too (cocos2d --> cocos2d/src and cocos2d/include), but in the spirit of keeping as much of the old structure as possible.

The .zip archives produced include a fat library and all the relevant headers with their original folder structure. Once a project is configured to use it then updating it is very trivial.

I have also enabled a lot more warnings, now treated as errors, and this has produced quite a few changes here and there in the codebase, but for the better I'd say :).
I did run cocos2d tests for both iOS and OS X targets (the former on Simulator only): particles, sprites, cocos2d-ui scrollview test.

CFLAGS with new warnings:
-Weverything -Wno-objc-missing-property-synthesis -Wno-pedantic -Wno-documentation -Wno-objc-interface-ivars -Wno-direct-ivar-access -Wno-vla -Wno-implicit-atomic-properties -Wno-padded

Note: Normally, I'd tend to go with the push against iVars from Objective-C (not the best thing for KVO), but this would be an even bigger refactoring that is not warranted at the moment.

Some warnings were suppressed on a per instance basis using the following kind of pragma (to make an example):
# pragma clang diagnostic push COCOS2D
# pragma clang diagnostic ignored "-Wfloat-equal"
# pragma clang diagnostic ignored "-Wignored-qualifiers"

[...]
# pragma clang diagnostic pop COCOS2D

In some cases, some floating point equalities need to be handled on a per-case basis.


Panajev:
The project is now back to the original structure and I have merged in  the latest changes.
I have fixed the two new static lib targets to adjust to this structure and the scripts behind them.

Birkemose:
It seems that several has required this?
Somebody at some point talked about allowing the script to install two versions of cocos2d. One as we know, and one with a static lib?

PR #344, opened 2013-10-05
CDAudioManager: interruption: method should not be compiled when targeting iOS < 6.0

wefiends:
This method calls beginInterruption: and endInterruption: both of which are only compiled on iOS > 6.0, so it won't compile if I target say, iOS 4.3. 

It's already commented as iOS 6.0 or greater only.


wefiends:
Looks like made the initial change on the develop-v2 branch by accident. Ignore all but the first commit.

PR #355, opened 2013-11-01
Update neon_matrix_impl.c

frranck:
Issue 1507: can't compile kazmath lib on arm64


Birkemose:
Can anyone who understands this, tell me if this is safe to merge?

dominik-hadl:
Don't know if safe, but kazmath doesn't compile on iPhone 5s and iPad Air if you have set up to compile for arm64 and this fixes it.

SiarheiFedartsou:
As far as I know in newest version of XCode(5.02) everything works fine without this fix.

PR #390, opened 2013-11-20
Port of CCLabel from cocos2d-x

SiarheiFedartsou:
It is a "stupid" port of CCLabel from cocos2d-x(via simple rewriting from C++ to Objective C).There are can be bugs, because I haven't time for normal testing.
Known problems:
1. Doesn't support ligatures(like cocos2d-x implementation). Do anybody know simple way to get ligature information from CTFont(for TTF fonts I've used CTFont as a backend)? Is ligatures support necessary?
2. Dynamic-generated font atlases can use more than one texture to store glyphs. CCSpriteBatchNode(superclass of CCLabel) doesn't support rendering with several textures.


Birkemose:
Due to everybody busy on the v3 release, I can not promise you a review this side of 1/1, but it definitely looks very promising.

PR #412, opened 2013-12-05
Added support for multitouch processing by grouping touches before calling CCResponder callbacks 

Ben-G:
**Multitouch capability for cocos2d**

In the current implementation on branch develop-v3, UITouches are always sent to CCResponders as single touches, even if multipleTouchEnabled is set to TRUE.

This happens because the current implementation iterates over all touches in the UIEvent and sends them immediately. This change implements a grouping & queueing of touches before in CCResponders before the touch method callbacks are called. 

With this implementation a CCResponder can check how many touches it received and determine if it is handling a single or multitouch.

Example implementation that uses my forked cocos2d: 
https://github.com/MakeGamesWithUs/Tutorial-Touch-Handling/tree/multitouch-cocos2d-test

_Please Note: calls to `- (void)touchesCancelled:(NSSet *)touches withEvent:(UIEvent *)event` in `CCResponderManager` are currently not grouped, since this would require larger changes in `CCResponderManager`. Suggestions how to solve this are welcome._

### DIFF
--- a/cocos2d-ios.xcodeproj/project.pbxproj
+++ b/cocos2d-ios.xcodeproj/project.pbxproj
@@ -560,6 +560,7 @@
 		50F9E8160E1AA34E000E7616 /* ccTypes.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; lineEnding = 0; path = ccTypes.h; sourceTree = "<group>"; };
 		50FBB2D8117613F500150761 /* CCActionTween.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = CCActionTween.h; sourceTree = "<group>"; };
 		50FBB2D9117613F500150761 /* CCActionTween.m */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.objc; path = CCActionTween.m; sourceTree = "<group>"; };
+		62EED1331850EE80008B17A7 /* CCResponder_Private.h */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; path = CCResponder_Private.h; sourceTree = "<group>"; };
 		6F305B5D0FB9D2790052E700 /* TransformUtils.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = TransformUtils.h; sourceTree = "<group>"; };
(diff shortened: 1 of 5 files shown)


Birkemose:
It is correct that touches are sent to nodes one at a time, to avoid to have to waste time on grouping them first.
There are no drawbacks or lack in functionality in multi touch, in doing it this way.

There currently is no functionality for counting how many touches a node currently processes, but I will see if I can add it before release.

Ben-G:
For the example code I have attached to the pull request it does have a drawback - in `touchesBegan` **I cannot detect if the user is starting a singleTouch or multiTouch interaction.** If my game's behaviour shall change depending on one or two finger touches I currently would have to build a workaround in the class that receives the touches. 

Additionally it may be a little confusing for people that have used UIKit before, where touches in a multitouch sequence are always sent together.

I've also been considered about performance at first, however all of the added code is only executed when `multiTouchEnabled` is set to true and multiple touches occur. In my first performance tests using profiler I couldn't detect any relevant performance influences - also not with multitouch interactions.

However, I understand if this is currently not a high priority. I also don't necessarily think that my suggested implementation is ideal, but I do think that implementing the touch handling analogous to UIKit would be great (and a lot of great changes in this direction have already been done - thanks).

Birkemose:
You are confusing terms. How many touches are in the set, has nothing to do with multi-touch or not.
If the user touches the screen with two fingers, 20mS apart, UIKit will also send two touchesBegan with one touch in each, even if it is a multi touch. Likewise, if you pinch, but only move one finger, touchesMoved will only contain one touch. 
In general, the touches set, only contains the changes since last.

I agree, that in in many cases there will be two touches, but relying on that, will result in missed multi touches, if you ex do not touch the screen at exactly the same time.

slembcke:
I was thinking about this the a couple weeks ago too. I thought it was odd that it sent sets with individual touches in them, but couldn't think of anything that it would cause problems with. UIKit's begin events almost always have a separate event for each finger since the the user's fingers have to be timed almost perfectly to arrive within the same event loop cycle. Touch move events also seem like they are so finicky about what is considered moving or not that the size of the set doesn't really tell you anything interesting. 

I'm actually really starting to wonder why UIKit sends them in sets at all instead of calling begin/move/end separately for each touch. Is there a good reason to keep using the sets within Cocos?

Birkemose:
There really is no good reason for using sets. I only did it to make the interface look similar to the original.

Birkemose:
I was wondering, if I should replace the sets with just the touch. To avoid confusion?

andykorth:
I'm a fan of that approach.

slembcke:
Yeah, I would go for it. I still cannot think of a reason to keep them in sets, especially if they are delivered one at a time anyway.

Ben-G:
Sounds good to me, too.

Issue #421, opened 2013-12-11
CCNode removeChild should assert when removing child that does not exist


andykorth:
Proposed change:
https://github.com/andykorth/cocos2d-iphone/commit/684234573de4167eb5526b3352beebb27a26b4da#diff-b132d6e5c0d197fdffbca133b6235978L704

Unit test for change:
https://github.com/andykorth/cocos2d-iphone/blob/develop-v3/UnitTests/CCNodeTests.m#L107

Current behavior was to silently continue without removing anything or any notification that the child is not present.

andykorth:
Change was merged in.

PR #422, opened 2013-12-11
Unit tests for CCNode, testing positionType, scaleType, conversions to Points and UIPoints, changing referencecorners, and removing children that do not exist.

andykorth:
Made many tests, have a few minor fixes, most of which were typos. Also fixed the issue described in https://github.com/cocos2d/cocos2d-iphone/issues/421

All unit tests are passing.


vlidholt:
Thanks, at a first glance looks good! I will review it tomorrow.

Issue #413, opened 2013-12-05
Default resolutions and CCAppDelegate class


vlidholt:
Currently the AppDelegate class is somewhat complicated, with lots and lots of options for setting up the CCDirector and GLView. I suggest that we create a CCAppDelegate class that contains all code that generally will not be modified by the user and let the user create a sub-class of that class for the configuration. We should also provide a simplified setup method.

By default, only the application:didFinishLaunchingWithOptions: method would be needed to be overridden. It could look something along these lines:

``` objc
- (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
{
    [self configureCocos2dWithPixelFormat: kEAGLColorFormatRGB565 screenMode:CCScreenModeFlexible animationInterval:1.0/60];
}
```

If power users really want to tweak the settings they can just implement the configure method themselves and do whatever they need there.

I'm thinking that we could provide two default 'screen modes', flexible and letter boxed. Flexible is the current default and a letter boxed screen mode would give developers a fixed logical coordinate system that will be scaled to fill the screen but crop off edges on some devices to keep the square aspect ratio.

I suggest that for the fixed size we use a logical coordinate space of 568 x 384 points. This may seem arbitrary, but would make it easy to do letter boxing on all iOS devices while keeping the 1x, 2x and 4x scaling (see picture below).

![resolutions](https://f.cloud.github.com/assets/1539812/1686740/a9146202-5deb-11e3-8aeb-b12c6fc9d97a.png)

dominik-hadl:
That's exactly what I was talking about in another thread - I would really like to see the app delegate setup a lot simplified (although I will probably override it each time :D). I just remember that when I first saw all the settings I was a little scared and wasn't sure what is all that code in there, so I think that newbies would be happy if it was easier to set it up.

cocojoe:
I like the idea of facade for this as there is rarely much for the average user to change, in recent memory I think the changes had been related to rotation compatibility in iOS6 , at least this way it would have been easy to roll out the fix.

Panajev:
The fixed resolution mode sounds nice, but it should allow the user to optionally set a frame texture... See Pocket Frogs on iPhone and iPad.

[web [iPhone msg]]

davilca:
Hello!

[...]

PR #424, opened 2013-12-12
Fixed screen size mode.

slembcke:
Basically done. Still not sure what to do about the viewSize problem mentioned in #413 though. If a dev is using the fixed size mode, there is no way of getting the correct screen size other than hard coding it. Maybe I should just expose the FIXED_SIZE variable and they just need to know to use that instead of CCDirector.viewSize? The default sizes of CCScenes and a few other nodes will be wrong too.

Also, it searches for the smallest power of two that will cover the screen. I'm not convinced this is the best solution, since it will pick a 4x content scale on a 1280x720 Android tablet and cut off a _lot_ of content (1/3 of the width and over half the height). Either the dev needs to ship the game with 3x art (terrible option), or we need to add some letterboxing to avoid severe cropping like this.

I did change some of the enums to NSStrings since this is intended to be used with an NSDictionary. Apple almost always uses strings for dictionary config parameters since wrapping enums in NSNumbers is a little tedious.

### DIFF
--- a/cocos2d/Platforms/iOS/CCAppDelegate.h
+++ b/cocos2d/Platforms/iOS/CCAppDelegate.h
@@ -56,7 +56,7 @@ NSString* const CCSetupTabletScale2X;
 /**
  *  Most Cocos2d apps should override the CCAppDelegate, it serves as the apps starting point. By the very least, the startScene method should be overridden to return the first scene the app should display. To further customize the behavior of Cocos2d, such as the screen mode of pixel format, override the applicaton:didFinishLaunchingWithOptions: method.
  */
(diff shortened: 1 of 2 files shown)


slembcke:
Oh, I forgot about screenshots. Here is the original image:
image

I just did the alignment by eye in photoshop, so it's not perfect.

It's set up to sit with it's lower left at the origin like so:

```
CCSprite *sprite = [CCSprite spriteWithImageNamed:@"ScreenSizes.png"];
sprite.anchorPoint = ccp(0, 0);
```

On iPhone:
iphone

On iPhone5:
iphone5

On iPad:
ipad

So they all pick the correct cropping area.

PR #426, opened 2013-12-12
CCColor Fixes

cocojoe:
Methods that call 'CCColor getRed:...' in iOS can't assume the Colorspace is RGB and will not work as expected (a few things will default to whiteColor which is in Mono Space).
In OSX this method should be fine however I refactored usage for consistency.

### DIFF
--- a/cocos2d/CCActionInterval.m
+++ b/cocos2d/CCActionInterval.m
@@ -1236,13 +1236,10 @@ static inline CGFloat bezierat( float a, float b, float c, float d, CCTime t )
 {
 	id<CCRGBAProtocol> tn = (id<CCRGBAProtocol>) _target;
     
-    CGFloat fr, fg, fb, fa;
-    [_from getRed:&fr green:&fg blue:&fb alpha:&fa];
+    ccColor4F fc = _from.ccColor4f;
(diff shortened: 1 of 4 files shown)


slembcke:
Ooof. Maybe we do just want to make our own color class that just wraps a float4 after all. I feel like it would probably take less than 30 minutes to rewrite it with a compatible API instead of wrapping NSColor/UIColor. Most of the code in there now could just be reused.

Birkemose:
Not saying “What did I say” … ;)

andykorth:
Yeah, after looking into the color code and writing these tests, I agree this could really be much better.
1. internal API uses a mix of different color types and structs.
2. In several places it uses a ccColor3 + float alpha, instead of a ccColor4.
3. Manual conversions between float and byte colors scattered everywhere through the code (already we found some mistakes were 1.0f was dumped into a byte color).
4. Personally I'm a fan of using ccColor4F internally everywhere. If having it compatible with UIColor is something people really like, I could understand that too. Although UIColor seems to have colorWithPatternImage, stroke vs. fill colors, which don't really apply. 

It would definitely go quickly, whatever route we pick.

PR #438, opened 2013-12-16
Rename color interpolate method, add docs

andykorth:


andykorth:
https://github.com/andykorth/cocos2d-iphone/commit/59ee7c090e433cabee1312b1128e8188f04d19b6 is part of a fix for SpriteBuilder, ease of use of NSColor

PR #462, opened 2013-12-30
Please review: Proposed changes to the templates and default CCFileUtils behavior.

slembcke:
Here are my proposed changes to the templates:
- It uses CCAppDelegate now so it's much simpler. I have a few example configurations shown. I tried to strike a balance of showing off some of the simple features and throwing a wall of text at the user.
- The old template was using the directory CCFileUtils mode. This was never the default before, and I'm not sure it's a good idea to change it now. If people want to change it then it's a must to at least add empty folders to the template. Otherwise it will be 100% unclear how to properly add assets without reading a lot of code first.
- CCFileUtils.enableiPhoneResourcesOniPad defaults to YES now. I really feel that this is expected behavior to many. The only conceivable downside is that it doesn't fail and let you know that you are accidentally loading an iPhone asset on an iPad.

As for the CCFileUtils changes. I strongly believe that we should be moving most users towards a simplified asset loading scheme. If we can pull off automatic downscaling for the next version (which I'm certain will be fairly easy even), then encouraging them to use suffixes will make it easier to transition.

### DIFF
--- a/cocos2d/Support/CCFileUtils.m
+++ b/cocos2d/Support/CCFileUtils.m
@@ -146,7 +146,7 @@ static CCFileUtils *fileUtils = nil;
 		
 		_bundle = [NSBundle mainBundle];
 
-		_enableiPhoneResourcesOniPad = NO;
+		_enableiPhoneResourcesOniPad = YES;
 		
 		_searchResolutionsOrder = [[NSMutableArray alloc] initWithCapacity:5];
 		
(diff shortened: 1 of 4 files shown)


cocojoe:
Sounds good to me, agree with CCFileUtils it should follow the 2.1 default which just worked and would be the expectation for any users upgrading from 2.

PR #467, opened 2014-01-03
Proposal: Use 'instancetype' as return type for factory methods

Ben-G:
This is the preferred return type for factory methods and increases type safety.

I used search->replace but checked that no warnings/errors were added during this refactoring.


Birkemose:
I think Apple recommends this, so I guess it should be changed.
What does others say?

cocojoe:
Well I thought about it when I was going through the header files however
once you start going down the road of 'this years best practice' you start
to ponder all sorts of changes.  Best start moving those private instance
variables into @implementation and dropping most of the implied @synthesize
;-)

Sounds fine to me, although always scary seeing so many files changed in
one go.

On 3 January 2014 14:45, Birkemose notifications@github.com wrote:

> I think Apple recommends this, so I guess it should be changed.
> What does others say?
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/cocos2d/cocos2d-iphone/pull/467#issuecomment-31526151
> .

## 

Martin

Ben-G:
Anyone want to make a final decision? :)

slembcke:
I don't see any reason why not. There will always be new language features to tidy up the code with: properties, blocks, fast enumeration, dynamic ivars, ARC, literals, indexing operators etc. Don't need to rush to use use each new feature, but no real reason to avoid them either.

Birkemose:
I think we should do it, just not now, so close to release. Furthermore, it can't automatically be merged.
What I think we should do afterwards, is in stead of renaming everything, we should go through the classes, and change them one by one. It is intimidating to have so many classes in a single merge. 
By doing them one by one, we also have a chance to fix the other things that Scott mentioned.

cocojoe:
Yeah agree with that, if we are going to do it let's agree on best practice
in general and apply per class.

On 10 January 2014 13:02, Birkemose notifications@github.com wrote:

> I think we should do it, just not now, so close to release. Furthermore,
> it can't automatically be merged.
> What I think we should do afterwards, is in stead of renaming everything,
> we should go through the classes, and change them one by one. It is
> intimidating to have so many classes in a single merge.
> By doing them one by one, we also have a chance to fix the other things
> that Scott mentioned.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/cocos2d/cocos2d-iphone/pull/467#issuecomment-32025413
> .

## 

Martin

slembcke:
Hah. I didn't mean that we had to change everything at once, but I'm always okay with cleanup. :p

Ben-G:
> slembcke  merged 0 commits into cocos2d:develop-v3  from Ben-G:develop-v3

I messed this one up, by not creating a separate branch for my pull-request -.-. If you're still interested, I'll provide the changes in a new pull request.

PR #469, opened 2014-01-05
iOS Template Changes

cocojoe:
Updated iOS template to do a bit more and actually do something a bit
more fun for the user. While covering a few of the key areas.


cocojoe:
As a heads up, @Birkemose is making a nice physics template.  So I think we should have two templates, the first being the classic sprite, action, touch template and the later letting the user get started with physics and covering delegate etc  I think this will be ideal for the end user.

Birkemose:
You can never have too many templates ...

I will send you my files tomorrow, then you can whip up the templates, and make a new pull request.

Issue #493, opened 2014-01-10
OSX templates can't link due to chipmunk


Birkemose:
I think Scott addressed this for IOS, is that correct?

slembcke:
Oh... I fixed it for the regular Mac projects, but I forgot about the templates. I'll look into it today.

dominik-hadl:
Fixed that in pull request #497 .

PR #497, opened 2014-01-10
OSX Template fixes

dominik-hadl:
This fixes some of the template problems (mainly on OSX) I've noticed. Project from OS X template now builds with only three warnings regarding deprecation of "Gestalt()", otherwise builds fine.


dominik-hadl:
Also fixes issue #493.

Birkemose:
Thanks :)

Issue #432, opened 2013-12-13
Installer + User Guide


Birkemose:
Maybe it is worth looking into if an installer / automator of some kind, could ease the process for newcomers? Nothing too fancy. Would be great if it could download latest version, have a version tracker of some sort, automate installation, and link to the forum (and maybe Rays site etc)

Since I did the hello world, maybe I should look into polishing that a little more for the release, and add a Users Guide to help a newcomer through the first couple of steps?

Any thoughts on this? Should a user guide be a PDF, or could it be integrated better with Xcode in some way?

davilca:
Hello people!

For new users, an installer and a quick start guide would be fantastic, because installation of the libraries were always a little annoying. For the future of cocos2d, and in order to grow the user base, it would be better to ease the installation process.

Personally, I prefer online guides or documents, instead of Xcode integration. In fact, it would be great to clean up and start from scratch the Documentation page (http://www.cocos2d-iphone.org/wiki/doku.php/) to fit the new workflow. I suppose that most of the new users get confused there, and that's why there are some recurrent forum post from time to time.

Also, I think that newcomers miss some more examples and use cases in the documentation. The cocos2d built in example projects are ok, but you must compile and execute them to see what they do, and then investigate the code to see how they do that. Maybe a good idea is to write code examples for the most important classes, showing images of the output result if needed. Also, you could link these examples to the class reference.

I suppose all this documentation clean up should be done when the API is stable.

cocojoe:
@Birkemose I don't think at this point we need some sort of cocos2d management application.  I think for now it would be sufficient to have a clear download section on site, with a download for cocos2d 3.0 DMG.

This could contain:
1) cocos2d installer - which would uncompress, run the install.sh and copy the new appledocs into Xcode. (I've started on this)
2) A web link to cocos2d community
3) A web link or 'getting started' PDF?

It is probably just easier to have a web link for the getting started section, this way it can always be current/updated.
A heavily documented, step by step 'hello physics world' tutorial would be good. Would be a great starting place.

It would of course be great to have a Sprite Kit style user guide but may take a bit of time to do this.

Now that chipmunk files are no longer built in the install script we can remove the requirement for CLI Xcode build tools to be installed.  
Should the chipmunk files then be include in the installer 'vs downloaded in script'?
I was also thinking to include pre compiled apple docs vs building in script (as need cli tools).

I would rather have the install 'just work' vs potential for various install issues...

[...]

PR #499, opened 2014-01-12
CLI Installer Build Script

cocojoe:
Issue #432
Also added a cli documentation builder, for use with installer.


Birkemose:
Merged

Edit:
Urk ... Looks like I messed up, so that the changes looks like mine. Sorry about that.

PR #513, opened 2014-01-20
Adds optional completion handler for actions

chrene:
This is just one thing I really have missed in the other versions. The ability to get notified when your actions are completed, using blocks.

I tested it with actions running simultaneously and the completion handlers were called as wanted.


Panajev:
Neat :)

[web [iPhone msg]]

dominik-hadl:
Very nice. Even though it was possible to chain a CCCallBlock in a CCSequence to achieve the same thing, this approach is much simpler.

Birkemose:
We are releasing 3.0 tomorrow, so everything is locked down for now. Looks like a really great feature, so someone will review it asap.

chrene:
Yeah saw the RC update on the forums, so that's what I thought too :)

cocojoe:
Can someone please review this.

Birkemose:
As far as I can see, the code looks solid and works.
When merged, we should definitely add some more action testing.

PR #520, opened 2014-01-22
Fixing a typo in comments in onEnter in two Scene files in the new project template

richardgroves:
Both NewtonScene.m and HelloWorldScene.m have 'Pr' rather than 'Per' in a comment in the onEnter method.


richardgroves:
Fixed some more documentation issues in CCTransition.h - two properties were not showing in the doc files as the documentation for them was rolled into the previous property in the header file. Made a mess with github - didn't branch before making this change, hence it being in the PR for the typos in the new project template.

Issue #521, opened 2014-01-23
Proposal: CCActionMove with flexible target


Ben-G:
I have prototyped a CCAction for one of our tutorials that allows a Node to follow a moving target until it reaches it. Would there be any interest in adding this to cocos2d? Or is it too much of a niche use case?

![actionfollowmoving](https://f.cloud.github.com/assets/1423068/1980880/f4b76a92-83ca-11e3-9557-217efa539098.gif)

If there's interest I'll polish it and create a pull-request. Once it reaches the target the action completes, that's why the penguin stays at the final position.

In the GIF you see my action combined with a rotation action. Sorry about it being sluggish.

cocojoe:
Yeah a CCActionHomingMissle ;) Would be good to have, I'm all for developer
convenience.

On 23 January 2014 01:15, Ben-G notifications@github.com wrote:

> I have prototyped a CCAction for one of our tutorials that allows a Node
> to follow a moving target until it reaches it. Would there be any interest
> in adding this to cocos2d? Or is it too much of a niche use case?
> 
> [image: actionfollowmoving]https://f.cloud.github.com/assets/1423068/1980880/f4b76a92-83ca-11e3-9557-217efa539098.gif
> 
> If there's interest I'll polish it and create a pull-request. Once it
> reaches the target the action completes, that's why the penguin stays at
> the final position.
> 
> In the GIF you see my action combined with a rotation action. Sorry about
> it being sluggish.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/cocos2d/cocos2d-iphone/issues/521
> .

## 

Martin

slembcke:
Closing this since the pull request for it was also closed.

PR #522, opened 2014-01-23
Added CCActionMoveToMovingTarget and Tests

Ben-G:
Adds the action mentioned in #521. Allows a target to move to a moving node or to a position that is provided by a block.

Optionally allows to provide a block performed on completion.
Optionally allows a node to follow the target position infinitely.

Preview of class interface:

```
  +(id) actionWithSpeed:(CGFloat)speed targetNode:(CCNode *)targetNode;
  +(id) actionWithSpeed:(CGFloat)speed targetNode:(CCNode *)targetNode followInfinite:(BOOL)infinite;

  +(id) actionWithSpeed:(CGFloat)speed positionUpdateBlock:(PositionUpdateBlock)block;
  +(id) actionWithSpeed:(CGFloat)speed positionUpdateBlock:(PositionUpdateBlock)block followInfinite:(BOOL)infinite;
```

Any suggestions for improvements welcome!


Ben-G:
Any feedback on this? Would love starting to use it in a tutorial.

cocojoe:
I liked the sound of this originally, not had any time to try yet, just skimmed code.  What was your tutorial use case that required the bottom 2 methods?

You could do all sorts with these really, for example a distance would allow you to create nice trail of nodes (I'm thinking R-Type).
Would it also be possible to trigger a block once upon reaching the destination of an 'infinite' based method, just thinking about a one time effect once the node reaches its target area.

On a trivial note I don't really like the name, it should probably be 'CCActionMoveToTarget' or 'CCActionFollowTarget' (although the later is like the infinite edition of this action :)

Ben-G:
> I liked the sound of this originally, not had any time to try yet, just skimmed code. What was your tutorial use case that required the bottom 2 methods?

In my case an object is not following the anchor point of a Sprite. It is following an arbitrary position within that Sprite (in this case it follows the scoop of the catapult, while the whole catapult arm is represented as one Sprite).

> Would it also be possible to trigger a block once upon reaching the destination of an 'infinite' based method, just thinking about a one time effect once the node reaches its target area.

Absolutely! I just couldn't come up with a clever API for this feature. But I could mention in the docs, that when applying a completion block on an infinite movement, it will only be called once.

> On a trivial note I don't really like the name, it should probably be 'CCActionMoveToTarget' or 'CCActionFollowTarget' (although the later is like the infinite edition of this action :)

Didn't like it either but failed to come up with a better one.

@cocojoe if you want to see the actions running you can take a short look at the tests I have written.

slembcke:
CCActionMoveToNode?

The only thing that seems odd to me about this is that the speed is relative to the parent's coordinate system and not world coordinates. Probably fine since parent nodes rarely have non-rigid transforms though.

Ben-G:
@slembcke renamed to 'CCActionMoveToNode'. Also added note that the speed is defined relative to parent's coordinate system. I added a test with a scaled parent node. For me the behaviour feels correct. Could you take a look at it? I'm happy to change the implementation if there's a better solution.

@cocojoe:

> Would it also be possible to trigger a block once upon reaching the destination of an 'infinite' based method, just thinking about a one time effect once the node reaches its target area.

I added this functionality.

vlidholt:
Although, this is well written and can in some cases be useful, I'm thinking it may be better as an extension to Cocos2d or even as a tutorial on how to write custom actions. It feels a little bit to special cased to be a part of the core Cocos2d (at least for my taste :P )

cocojoe:
Perhaps we should introduce an extension sub module for cool stuff, which
is fun to have but should not distract from core development. I have a few
actions in mind that would fit nicely into this.  The only downside I've
seen in the past with such extensions is they often become unmaintained.

On 29 January 2014 22:13, Viktor Lidholt notifications@github.com wrote:

> Although, this is well written and can in some cases be useful, I'm
> thinking it may be better as an extension to Cocos2d or even as a tutorial
> on how to write custom actions. It feels a little bit to special cased to
> be a part of the core Cocos2d (at least for my taste :P )
> 
> ## 
> 
> Reply to this email directly or view it on GitHubhttps://github.com/cocos2d/cocos2d-iphone/pull/522#issuecomment-33637485
> .

## 

Martin

Ben-G:
@vlidholt Yes, that was my concern at first. I probably agree that this works better as an extension ;)

PR #542, opened 2014-01-28
When the TextureTest is run and exited back to the list of tests the file search paths are left incorrect

richardgroves:
The setup method of the TextureTest purges the shared texture cache and the sets the CCFIleUtils search paths to "Images" and the DefaultSearchPath (which is ""). On exiting the test it does not restore the previous setting of the search paths.

This causes problems:
1) the search paths are initially "Resources-shared" and "" - so once the paths are changed the UI interface files can't be loaded
2) Once the test exits no other tests can load/access things like the Interface.png texture

The fix:
1) Pre-pend the "Images" path to the existing search paths
2) Remove the "Images" path on exiting the test (pressedBack method)

NB: The ParticleTest and ColorTest also do the same overwrite of the search paths but suffer less bad effects as they don't purge the texture cache. They should still also be fixed to correct the problem I think.

### DIFF
--- a/cocos2d-ui-tests/tests/TextureTest.m
+++ b/cocos2d-ui-tests/tests/TextureTest.m
@@ -22,7 +22,19 @@
 {
 	[self.contentNode removeAllChildren];
 	[CCTextureCache purgeSharedTextureCache];
-	[[CCFileUtils sharedFileUtils] setSearchPath: @[ @"Images", kCCFileUtilsDefaultSearchPath] ];
+	NSMutableArray *newSearchPaths = [[[CCFileUtils sharedFileUtils] searchPath] mutableCopy];
+	[newSearchPaths insertObject:@"Images" atIndex:0];
+	[[CCFileUtils sharedFileUtils] setSearchPath: newSearchPaths ];
+}
+
+- (void) pressedBack:(id)sender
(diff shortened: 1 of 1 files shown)


andykorth:
Ah, good catch Richard! One suggestion is to use:

```
[CCFileUtils resetSingleton];
```

Which will reset _all_ the state tracked by CCFileUtils. That approach is used in CCFileUtilTests.m, which makes many changes to search paths.

richardgroves:
But wont 'resetSingleton' clear the search paths back to default - ie "" - when the AppDelegate for the tests project has set them up to include where it finds its interface elements?

It's fair enough to do that in a unit test where a clean starting point is wanted, but the interactive tests are a bit different I think. 

Hmmm....just did a quick test and resetSingleton does set the search path to just "", but it continues to work for a while as the interface elements are reloaded during the TextureTests so are in the cache after it exits - but if you go back into the TextureTests once it purges the cache on setup it has lost a way of finding the interface elements and gets messed up.

andykorth:
Oh, I didn't realize the interactive testbed itself was modifying the search path. Sounds good, thanks for explaining it. I think I confused myself with the unit test behavior vs. the interactive tests.

PR #544, opened 2014-01-28
fixes memory leak in CCActionManager

Ben-G:
Fixes a memory leak in CCActionManager. With the old implementation, if the _actions_ array already existed a NSMutableArray would be initialized and retained but the pointer to this array would not be assigned to anyone, resulting in a dangling retained object and a memory leak.

### DIFF
--- a/cocos2d/CCActionManager.m
+++ b/cocos2d/CCActionManager.m
@@ -81,13 +81,13 @@
 
 -(void) actionAllocWithHashElement:(tHashElement*)element
 {
-    NSMutableArray* aObj = [[NSMutableArray alloc] init];
-    void* a = (__bridge void*) aObj;
-    CFRetain(a);
-    
 	// 4 actions per Node by default
-	if( element->actions == nil )
+	if( element->actions == nil ) {
+        NSMutableArray* aObj = [[NSMutableArray alloc] init];
+        void* a = (__bridge void*) aObj;
+        CFRetain(a);
 		element->actions = aObj;
+    }
 }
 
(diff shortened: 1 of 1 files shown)


richardgroves:
Hi Ben,
Is there a good reason that the _actions_ array is ___unsafe_unretained_? Removing that would simplify the code, or is there something implicit in the HASH_\* macros that require it?

Ben-G:
@richardgroves The array is stored in a struct and ___unsafe_unretained_ is the only supported memory type (ARC cannot handle memory management for objects stored in structs: http://stackoverflow.com/questions/8093099/arc-forbids-objective-c-objects-in-structs-or-unions-despite-marking-the-file-f)

richardgroves:
Thanks - I've a vague memory of reading that when learning about ARC conversion, but had never seen it in the wild.

Is there a case then for making tHashElement an Objective-C object? It would make the system more 'objective-C' and I, at least, always get a bit nervous when I see the casts + CF... calls for doing manual memory management.

[Probably dragging what was a nice memory leak fix on too long here - just curious about how far the 'clean up' of old code to native Objective-C is going]

slembcke:
Ah, good catch. Merged.

Ben-G:
@slembcke "Closed with unmerged commits" <- somehow the changes don't seem merged?

slembcke:
Err, what? How did that happen?

edit: Oh... I'm willing to bet I clicked the merge button, didn't confirm it and then hit the comment & close button. Derp.

PR #557, opened 2014-01-30
GLKit migration

SiarheiFedartsou:
Added CCMath module with code that repeated kazmath functionality(that was using in cocos2d) using GLKit API.


slembcke:
Uh-oh. I already made the GLKit switch in the renderer branch. :-\ I probably should have commented on that, sorry.

SiarheiFedartsou:
:(

PR #562, opened 2014-02-03
Adding Ease Sine actions

iMobCoding:
I already asked on cocos2d forum, but without response. I find those EaseSine actions very useful, so I added them again...

http://www.cocos2d-iphone.org/forums/topic/any-reason-to-drop-easesine-actions-from-v3/

### DIFF
--- a/cocos2d/CCActionEase.h
+++ b/cocos2d/CCActionEase.h
@@ -59,6 +59,31 @@
 @end
 
 
+#pragma mark - Ease Sine Actions
+/**
+ *  This action will start the specified action with an sine effect.
+ *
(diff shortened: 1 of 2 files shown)


KingOfTheHuns:
+1 for this. I think also, SpriteBuilder should support EaseSin actions when creating animations.

cocojoe:
Makes sense.

On 3 February 2014 13:15, KingOfTheHuns notifications@github.com wrote:

> +1 for this. I think also, SpriteBuilder should support EaseSin actions
> when creating animations.
> 
> ## 
> 
> Reply to this email directly or view it on GitHubhttps://github.com/cocos2d/cocos2d-iphone/pull/562#issuecomment-33952970
> .

## 

Martin

PR #563, opened 2014-02-03
CCLayoutBox incorrectly positions children in vertical direction

iMobCoding:
Children are shown in reverse order when added to vertical layout box. With horizontal it works as expected.

### DIFF
--- a/cocos2d/CCLayoutBox.m
+++ b/cocos2d/CCLayoutBox.m
@@ -84,7 +84,7 @@ static float roundUpToEven(float f)
         
         // Position the nodes
         float height = 0;
-        for (CCNode* child in self.children)
+        for (CCNode* child in [[self.children reverseObjectEnumerator] allObjects])
         {
             CGSize childSize = child.contentSizeInPoints;
             


slembcke:
Not really sure I'd consider this an issue. Though perhaps it's more useful in general to have objects arranged in descending order, I'm not sure if it's really wrong to have them arranged with increasing coordinates.

Anybody else have thoughts?

andykorth:
Zero is the bottom of the screen, that's why the first item appears at the bottom. _shrugs_

iMobCoding:
I am sure it's a copy/paste bug from the horizontal layout box, whose code is just above the vertical one... For horizontal you add child 1, 2, 3... and it shows child 1, 2, 3... from left to right. So it's logical when you add child 1, 2, 3... to vertical layout, to show child 1, 2, 3... from top to bottom also. I just don't see a reason for it to arrange children from bottom to top. Anyway, just check how vertical CCMenu added items in v2...

Best

Birkemose:
I completely agree. If adding 3 items to a menu, first item added should be at the top.

vlidholt:
Not exactly a bug, but from a usability perspective using the top-to-bottom ordering may make more sense, I'm merging this.

iMobCoding:
When you expect to release new version of SpriteBuilder, because right now in cocos it's top-down but in SprBuilder is still bottom-top order?

Best

vlidholt:
Hm, didn't think of this. Maybe we should revert this change, or otherwise we will need to write some sort of patch to SpriteBuilder that reverse the order of nodes in a layout to keep things backwards compatible...

vlidholt:
A new version of SpriteBuilder is planned closely after the release of Cocos2d 3.0 final.

iMobCoding:
Why don't you release v1.0.5 in the meantime with latest cocos commit, it's labeled RC2 anyway...

PR #581, opened 2014-02-10
64 bit fixes.

slembcke:
Finally picked up a 64 bit iOS device last week and found out there were some problems in both Chipmunk and Cocos2D. I was sort of surprised since both have been running on 64 bit Intel for a while. There were actually a few fairly serious bugs.

I figured these changes should get reviewed by at least one other person before merging:
- Fix all the warnings when building for 64 bit.
- Update the deployment target to 5.1.1 as required by the 64 bit arch.
- Make sure there are casts to all objc_msgSend() and method imp calls.
- Fixed a few bad pointer casts (GLfloat \* to CGPoint *, etc)
- Updates the Chipmunk reference where I fixed a number of similar issues there.
- Make it optional to run the physics using float32 on 64 bit for a mild, but reasonable performance bump. Most of the diffs are for this. (defaults are still to use CG types for convenience)
- In general, make a better separation between CG and Chipmunk types.


jjxtra:
Please let's get this merged in ASAP

dominik-hadl:
+1 I would really appreciate this.

Birkemose:
I have gone though the changes, and it all seems good to me. All tests seems to work on a 64 bit target.
Seems that target was not set to 64 bit, and min ios version has to be upped to 5.1.1 (ircc)
There is a couple of compiler and analyzer warnings, but nothing seems to be related to 64 bit.

I would say it is go.

PR #585, opened 2014-02-12
Add CCSlider to the UI headers

masanorythm:


slembcke:
Weird. The header was explicitly included in the slider tests. Somebody must have missed it.

PR #588, opened 2014-02-13
Fixing a leak in CCControl

dominik-hadl:
This was the cause of a crash if the you repeatedly pressed a CCButton in short succession. The target wasn't deallocated and it crashed with EXC_BAD_ACCESS (code 1), which is an invalid pointer. This should fix it and since the deployment target is >= 5.0 there's no reason for it to be unsafe unretained.

### DIFF
--- a/cocos2d-ui/CCControl.m
+++ b/cocos2d-ui/CCControl.m
@@ -59,7 +59,7 @@
 
 - (void) setTarget:(id)target selector:(SEL)selector
 {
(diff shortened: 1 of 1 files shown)


slembcke:
You know, I just looked at those when I was going through the objc_msgSend() call for 64 bit. I looked at them suspiciously, then left them alone. There are a couple more of them I should track down as well.

PR #589, opened 2014-02-13
Redraw visible dirty rows.

toreolsensan:
This prevents visible cells from disappearing when you call [tableView reloadData].


toreolsensan:
I'm closing this, since @masanorythm's fix in #603 seems to solve the same problem in a much simpler way.

PR #595, opened 2014-02-15
Remove content scale from bitmap font label as it was causing labels to be way too far down

jjxtra:
bmGlyph app with align to grid, bitmap fonts work great with this change.


lgoss007:
I was having all kinds of issues trying to line up my bitmap fonts on the bottom of the screen. Different devices would put it in different places. This fixed the issues.

jjxtra:
Glad it's working for you. This doesn't fix everything if align to grid is off, but at least bitmap fonts are usable with it.

cocojoe:
@jjxtra or @lgoss007 can you provide a sample project of bitmap fonts behaving badly.  I want to add a few more tests cases.  Thanks

jjxtra:
Sure thing, hang on a few minutes...

jjxtra:
I've got a sample project and notice two AppDelegate options that cause things to have problems:
CCSetupScreenMode: CCScreenModeFlexible,
and
CCSetupTabletScale2X: @(YES)

With these options commented out, the bitmap font works fine. The pull request change to the content scale allows the bitmap font to work even with these options turned on.

Try the sample project on non-retina iPad and then retina iPad to notice how bad it gets on Retina.

Sample project: https://www.dropbox.com/s/lti0qsdnpi2cul7/BitmapFontTest.zip

cocojoe:
Thanks @jjxtra 
My gut feeling was this probably fixes it for you but breaks it for someone else, so I ran through some tests with other fonts.

ipad_retina_default
ipad_retina_default_patched
ipad_retina_tabletscaleon
ipad_retina_tabletscaleon_patched

I can't accept this pull however i'll raise up the issue in general with the given examples.

slembcke:
It just occurred to me that the issue is probably that it should be using the texture's content scale instead of the screen's. I'll check it out tomorrow maybe.

@cocojoe, what are the non-working fonts? I'd like to add them to the UI tests.

cocojoe:
@slembcke Good idea, just replaced __ccContentScaleFactor with _textureAtlas.texture.contentScale it all 'just works' now. 

You can grab the retro font from the project above, the other one is in a commercial project but tbh this probably effects most fonts.

I'll do a pull later.

cocojoe:
Just to follow this up you don't really need any other fonts, you can create this with the current unit test font by simply removing say the iPad Retina resource of the font and running on iPad Retina.
So if you have all available resolution resources then this issue will not appear, it's more on fallback resolutions being used.

slembcke:
Ah, gotcha. I'll close this one and merge #612 then.

PR #603, opened 2014-02-19
Fix CCTableView reloadData

masanorythm:
[CCTableView reloadData] is not working redraw
Has reset the CurrentlyVisibleRange.


vlidholt:
Thanks!

Issue #598, opened 2014-02-17
Z-Order, Vertex-Z and Sprite Batch Nodes don't work


davilca:
I know this issue was solved weeks ago. I also have looked into the vertex-z test case added by Birkemose, but I couldn't run it (due to some Chipmunk issues).

I have created a fresh clean project using the V3 template. You can dowload the test project here:

Test Project: http://dyabitgames.com/files/prueba.zip

In this test, 3 sprites have been added to the batch node, one after another, they look correct.

Another 3 sprites have been added one after another, but I have reversed their drawing order using vertex Z. The result is incorrect. The transparency is not working. The transparent pixels appear as solid color (the background color) and cover the other sprites.

![vertexz-batch-nodes](https://f.cloud.github.com/assets/371049/2187916/7c68e43e-97fd-11e3-92d9-1653a483ed8b.png)

In the app delegate, I have used several depth formats, like this:

depthFormat: GL_DEPTH24_STENCIL8_OES

Why is this happening? This is a fresh clean project, with a batch node and some sprites.

Thanks in advance.

davilca:
People, Z-Order reordering isn't working neither. I have updated the Test Project to demonstrate that.

Test Project: http://dyabitgames.com/files/prueba.zip

![captura de pantalla 2014-02-18 a la s 14 57 40](https://f.cloud.github.com/assets/371049/2195720/acf2f02a-98a4-11e3-89fd-1a13a6306dff.png)

The program will reverse Z-Order and Vertex-Z each 2 seconds, with these results:
1. Z-Order is never changed (so [sprite setZOrder:zOrder] doesn't work??).
2. VertexZ is changed, but the rendering is incorrect. Also, it doesn't work for negative vertex-z values.

People, all this worked in v2.1, so am I doing something wrong or is it a bug?

cocojoe:
Have tested against #605 #604 This fixes Z-Order.

PR #604, opened 2014-02-19
updated setZOrder: in CCNode.m

kalo:
setting the zOrder of the current node is redundant if parent is present and will directly interfere with if( z == child.zOrder ) check in reorderChild:z: from CCSprite, CCSpriteBatchNode and CCParticleBatchNode causing them to defer from reordering their children. 

this is related to issue #598.


slembcke:
Thanks!

PR #605, opened 2014-02-19
updated swap:withNewIndex: in CCSpriteBatchNode.m

kalo:
textureAtlas quads are not updated to reflect changes to the atlas index of a swapped object. in consequence the draw order does not change. 

related to issue #598.


slembcke:
Thanks!

Issue #560, opened 2014-02-02
Particles are paused after popscene


leszek-s:
I have 2 scenes. On the first scene (main menu) there are some flying particles. If I push another scene and then pop back to the first main menu scene then particles on the main menu should continue to fly but they are paused in position that they were previously before switching the scene.

EDIT: also when I have - (void)update:(CCTime)dt implemented then it will also not be called after popscene.

slembcke:
There were a couple pause related scheduling bugs that were fixed in the last week or so. Would you be able to grab the latest git version and see if it was fixed?

leszek-s:
I just checked with the latest version and the problem still exists. Particles are paused and update is not called.

cocojoe:
This will be related to #608 I'm sure.

Are you pushing with a transition?

Issue #616, opened 2014-02-23
CCPhysicsDelegate not fully restored on Push/Pop scene


cocojoe:
If you push then pop a scene that includes a physics collision handler it will not be restored to existing scene physics objects.  New physics objects will however be handled by the collision handler.  Existing physics objects will still behave as expected in the simulation.

https://www.dropbox.com/s/n0e3k2y3dt3pbog/Cocos2DSimpleGame.zip

Issue #608, opened 2014-02-19
After calling popScene update is no longer called


TheProphetOfRa:
I have two scenes, the first one triggers the second with pushScene: withTransition: and the second one falls back to the first with popScene. After popScene has been called the first scene's update method is no longer called. popScene: withTransition has also been tried neither with any success.

slembcke:
Sounds like the scene isn't getting unpaused. Possibly related to this issue https://github.com/cocos2d/cocos2d-iphone/issues/608.

MighteeCactus:
Facing same issue. It happens only when you use push with transition. Without transition everything is fine. 
After calling push without transition and then pop with/without transition everything works properly.
I found that if you have simple node graph i.e. animated node is a child of the scene node, than something like:

```
- (void)onEnter
{
    [super onEnter];
    [self setPaused:NO];
}
```

can resume the animation after push/pop transition, but if you have smth more complicate than that transition failed to resume popped scene...
Overriding 

```
[node onEnterTransitionDidFinish] 
```

won't help either. Btw it is called BEFORE the 'paused' property will be changed last time by the transition...

I hope this will help.

cocojoe:
@MighteeCactus please can you provide your more complicated test project please.

cocojoe:
After looking into this, when a scene is pushed with a transition it will become paused.  
There is nothing to resume it with a direct pop, a transition will try and set the scene's pause flag but it doesn't really help.

I'll do a pull with the fixes shortly.

PR #617, opened 2014-02-24
Ensure pushed scenes directly or via transition will resume with pop correctly.

cocojoe:
Can someone with more CCDirector experience check this please.
Fixes: #560 #608 #616


Birkemose:
The problem as I see it, is, that is the scene was paused before pushing, it will be unpaused when popping.
I havent looked into the original problem though.

cocojoe:
If there is a transition with the push it will pause the outgoing scene.  A subsequent pop will not resume it.

A straight push/pop works without transition worked fine.

Birkemose:
Let me check up on it

Birkemose:
We will need an extra pause variable, if we want to preserve user pause when popping a scene. For now let us merge this, and simply force pause off when the scene is popped.

slembcke:
I agree that it's an improvement, but still not a complete fix. It seems like a pretty minor bug that somebody would pause a scene and then push it though. I guess we should merge it as is and make a separate issue?

slembcke:
I thought this got merged. :-\ I guess I'll push the button.

PR #636, opened 2014-03-07
Fix outline color on CCLabelTTF.

toreolsensan:
Without this, setting a CCLabelTTF's outlineColor property has no effect.


nheagy:
Beautiful, I have been wondering what was up with that.

vlidholt:
Has this been tested with both iOS 6 and 7? Does setting the width of the outline and shadow work correctly?

toreolsensan:
@vlidholt This has been tested on devices running iOS 6.1.3 and 7.0.6. Shadows with different color/offset works, as does the outline width (it seems NSAttributedString doesn't explicitly set line width, like it does with the foreground color). shadowBlurRadius does work. Are there any other parameters you think should be checked?

gimsangtae:
Thanks. great work.

toreolsensan:
@slembcke What happened to this pull request? Suddenly it contains thousands of commits, and was apparently closed without merging my commits. Is there any point in resubmitting?

orionnoir:
check out some of the other recent posts. turns out there might have been
some very unintentional side effects of a recent action doing something for
3.0. I haven't seen any resolution decision yet, but I read it was
unintentional.

slembcke:
It was closed automatically as an unintentional side effect. when we tried to cleanup and rename the branches. More details here:
http://www.cocos2d-iphone.org/forums/topic/important-read-me-git-users/

PR #639, opened 2014-03-07
Re-add CCAnimation.h and CCAnimationCache.h to cocos2d.h

nheagy:
As the replacement CCAnimationNode hasn't made it into 3.0, these ought to be back in the main header.

### DIFF
--- a/cocos2d/cocos2d.h
+++ b/cocos2d/cocos2d.h
@@ -57,6 +57,8 @@
 #import "CCActionInterval.h"
 #import "CCActionProgressTimer.h"
 #import "CCActionTween.h"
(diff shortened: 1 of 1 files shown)


vlidholt:
These classes are removed for a reason. They will not be supported in 3.1, instead we will be adding a new CCAnimation class with much more features than the old one. Re-adding these files would break future backwards compatibility.

You can still use them now by including the header files, but as they are part of the private API this will alert users that these classes are subject to change in the future.

nheagy:
"break future backwards compatibility"

So you're saying there is no officially supported sprite animation code in 3.0? I'm gobsmacked.

PR #665, opened 2014-03-29
Added objc_requires_super attribute to onEnter/onExit CCNode methods

lucianomarisi:
This saved me a lot if time with finding subtle bugs. I have not used the Objective C macro NS_REQUIRES_SUPER because builds with apportable break.


slembcke:
!! I did not know that existed. Apparently it's been around for a while. I think every Cocos programmer has messed that up a dozen times.

I'll email the Apportable devs about the macro version.

PR #667, opened 2014-03-31
Fix outline color on CCLabelTTF

toreolsensan:
Without this, the outline would always be black.

Tested on devices running iOS 6.1.3 and 7.0.6. Shadows with different color/offset works, as does the outline width (it seems NSAttributedString doesn't explicitly set line width, like it does with the foreground color). shadowBlurRadius does work.

Note: This is a resubmission of pull request #636 which was closed during the apocalypse.

### DIFF
--- a/cocos2d/CCLabelTTF.m
+++ b/cocos2d/CCLabelTTF.m
@@ -650,10 +650,9 @@ static __strong NSMutableDictionary* ccLabelTTF_registeredFonts;
         CGContextSetTextDrawingMode(context, kCGTextFillStroke);
         CGContextSetLineWidth(context, outlineWidth * 2);
         CGContextSetLineJoin(context, kCGLineJoinRound);
-        CGContextSetStrokeColorWithColor(context, [color CGColor]);
-        
+
         NSMutableAttributedString* outlineString = [attributedString mutableCopy];
-        [outlineString removeAttribute:NSForegroundColorAttributeName range:NSMakeRange(0, outlineString.length)];
+        [outlineString addAttribute:NSForegroundColorAttributeName value:color range:NSMakeRange(0, outlineString.length)];
         
         [outlineString drawInRect:drawArea];
         


slembcke:
Seems good. Sorry this sat around until the repo-calypse happened. :-\

vlidholt:
This is unfortunately not a correct solution. See issue: https://github.com/spritebuilder/SpriteBuilder/issues/271

toreolsensan:
@vlidholt Maybe my vision or my monitor is bad, but in the screenshot you posted in the spritebuilder issue I have a hard time seeing the difference. Can you please explain what exactly the problem introduced by my change is?

For the record, without this fix I always see the outline color as black _both_ on iOS 6 (6.1.3 at least) and 7. You claim the old approach was working for iOS 6, something I'm unable to verify.

slembcke:
It's subtle, but he's right. Each character is being shadowed slightly by the next character to the right. (Most visible to me on the "uil" sequence)

Is this worse or better than before? Should I revert the merge? Does anybody have time to set up some test cases in the ui-tests app?

toreolsensan:
I see the difference now. I was looking too hard on the lower shadow. I'm creating a test case for this now which we can use to discuss further.

toreolsensan:
@vlidholt @slembcke I made a test case: https://gist.github.com/8b10c17df3c16ebd39bc.git. If this is helpful I can submit it as a pull request. Can we use this to identify a solution?

NickyWeber:
Should be fixed in pull request Bugfix: Shadows overlapping/glow in stroke. #679.

vlidholt:
I've merged Nicky's pull request and it all looks good now as far as I can tell! :) Good work both @toreolsensan & @NickyWeber thanks, really happy we got this fixed!

huabin:
@toreolsensan @vlidholt  I think this bug still remains in V3.0, it's not fixed. btw @toreolsensan can you repair this test case link ( https://gist.github.com/8b10c17df3c16ebd39bc.git), it's down.

PR #668, opened 2014-03-31
Bitmap font label performance improvements

jjxtra:
Also contains a fix for rare cases where animation remained off when un-backgrounding an app


slembcke:
Merged this and then nearly immediately found another bug in the BMFont code that snuck in a few months ago when switching from tags to names.

I ended up reverting some of the code to a much earlier revision and making an array to map the tags to children. Avoids an O(n) lookup as well.

jjxtra:
Groovy. Thanks for pulling it in. As long as we are not doing NSString stringWithFormat, that's where the performance problem was :)

PR #671, opened 2014-04-01
Add popToRootSceneWithTransition: to CCDirector.

sdrew:
Adds a version of popToRootScene that takes a CCTransition object, to match the `WithTransition:` versions of `popScene`, `presentScene`, `pushScene` and `replaceScene`.


slembcke:
Makes sense and is straightforward.

PR #673, opened 2014-04-02
CCBAnimationManager Redux

cocojoe:
(Best have Viktor review this)

Refactored animation playback to create keyframe action sequence on demand between any given keyframes
 PlaybackSpeed Property
 Uses scheduler to control actions independently from CCActionManager
 Time Seek methods 
 Manual simple sequence builder (e.g. frame animations)
 Cocos2D foundation CCAnimation style wrapper
 Refactored ivar naming inline with Cocos2D v3
 Keyframe caching for extra performance
 Reverse animation code (negative playback speed)

CCAction*
 Refactored custom CCBActions into CCActions 
 Fixed rotational skewing x/y bug when running in parallel
 Modified CCRotateTo to allow for a simple rotation without overthinking it. (SpriteBuilder compatible)
 Added additional copy methods for CCRotateTo
 Added copy methods for refactored ccbactions

### DIFF
--- a/cocos2d-ui/CCBReader/CCBAnimationManager.h
+++ b/cocos2d-ui/CCBReader/CCBAnimationManager.h
@@ -25,105 +25,90 @@
 
 #import <Foundation/Foundation.h>
 #import "cocos2d.h"
+#import "CCBSequenceProperty.h"
 
 @class CCBSequence;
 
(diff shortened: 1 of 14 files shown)


cocojoe:
Urgh, wrong branch.

PR #675, opened 2014-04-02
Avoid a heap alloc of NSUInteger in [CCNode visit]

jhurliman:
While profiling a Cocos2d project I noticed that [CCNode visit] was allocating and deallocating an object every call. Some testing showed that changing the NSUInteger to an int removed this alloc/dealloc and resulted in a minor speedup for that method.

### DIFF
--- a/cocos2d/CCNode.m
+++ b/cocos2d/CCNode.m
@@ -898,7 +898,7 @@ RecursivelyIncrementPausedAncestors(CCNode *node, int increment)
 
 		[self sortAllChildren];
 
-		NSUInteger i = 0;
+		int i = 0;
 
 		// draw children zOrder < 0
 		for( ; i < _children.count; i++ ) {


slembcke:
Say what? NSUInteger is not an object. It's just a type alias for an `unsigned int` (maybe `unsigned long` on 64 bit? I forget.)

Either way, it's just a regular integer type and no allocation should be taking place. Can you post the part of you trace that made you think otherwise?

jhurliman:
Ah, you're right. I'm not sure what is causing the allocation in that method then, or why I saw it disappear. The only other change I made (not in the pull request) was switching to ObjC literals for the array access (ie `CCNode *child = _children[i];`) which should have no change. I attached a screenshot showing the objc_retain and objc_release calls happening inside [CCNode visit].

screen shot 2014-04-02 at 11 08 05 am

slembcke:
You can't avoid that with ARC. It's a bit more clear if you look at the assembly, but pretty much every return value or argument to a function is retained and released. It's really frustrating since it can add up to a lot of CPU time doing nothing. Switching to NSFastEnumeration is a good way to avoid this. I did that and a number of other optimizations to the visit method in the v3.1 code. It's crazy that ARC can make that method blow up so badly that it shows up in a time profile.

https://github.com/cocos2d/cocos2d-iphone/blob/develop/cocos2d/CCNode.m#L880

jhurliman:
Thanks for the explanation. That's good to know for optimizations in our own app as well. Looking forward to v3.1 too!

Issue #660, opened 2014-03-26
CCScheduler can't paused for the first time


Osnobel:
I added a schedule like 

```
[self schedule:@selector(addDrop:) interval:1.f repeat:CCTimerRepeatForever delay:1.5f];
```

But it can't paused for the first time. I  can see the new CCScheduledTarget is implicitly paused, but the new CCTimer not.

```
-(void)setPaused:(BOOL)paused
{
    if(paused != _paused){
        FOREACH_TIMER(self, timer) timer.paused = paused;
        _paused = paused;
    }
}
```

The paused == _paused == Yes when set  CCScheduledTarget.paused as "Yes" for the first time, so it can't paused for timers.

PR #676, opened 2014-04-02
CCNode setPaused different behavior proposal...

iMobCoding:
I have setup with **gameScene** (self) and animated **pauseLayer** as a child. With current cocos2d version if I do:

``` Obj-C
self.paused = YES;
self.pauseLayer.paused = NO;
```

animated _pauseLayer_ will not show, because its parent is already paused and _pauseLayer_ will not change its paused status. But if I remove **if statement** from the code, everything works as expected. So I have more control over which children are paused and which are not. I don't know if removing those lines have other implications, so please somebody else test it.

I know somebody suggested other approaches, like adding all game content to _self.someNode_ and then pausing only _someNode_, which works great, except I also want to pause all other schedulers and actions that are run on self and it's impossible this way, as self in that case isn't paused...

Best

### DIFF
--- a/cocos2d-ios.xcodeproj/project.pbxproj
+++ b/cocos2d-ios.xcodeproj/project.pbxproj
@@ -524,9 +524,6 @@
 		A06BF84C168109170018A802 /* libsqlite3.dylib */ = {isa = PBXFileReference; lastKnownFileType = "compiled.mach-o.dylib"; name = libsqlite3.dylib; path = usr/lib/libsqlite3.dylib; sourceTree = SDKROOT; };
 		A06D9C131728D5F600704732 /* CoreText.framework */ = {isa = PBXFileReference; lastKnownFileType = wrapper.framework; name = CoreText.framework; path = System/Library/Frameworks/CoreText.framework; sourceTree = SDKROOT; };
 		A0707F6215F946E9001C1263 /* LICENSE_CCBReader.txt */ = {isa = PBXFileReference; lastKnownFileType = text; path = LICENSE_CCBReader.txt; sourceTree = "<group>"; };
-		A0707F6315F946E9001C1263 /* LICENSE_JRSwizzle.txt */ = {isa = PBXFileReference; lastKnownFileType = text; path = LICENSE_JRSwizzle.txt; sourceTree = "<group>"; };
-		A0707F6415F946E9001C1263 /* LICENSE_jsbindings.txt */ = {isa = PBXFileReference; lastKnownFileType = text; path = LICENSE_jsbindings.txt; sourceTree = "<group>"; };
-		A0707F6515F946E9001C1263 /* LICENSE_SpiderMonkey.txt */ = {isa = PBXFileReference; lastKnownFileType = text; path = LICENSE_SpiderMonkey.txt; sourceTree = "<group>"; };
 		A07EC9B2141A8D32008C1F12 /* README.md */ = {isa = PBXFileReference; lastKnownFileType = text; path = README.md; sourceTree = "<group>"; };
 		A08C378E14144FBB004538FD /* neon_matrix_impl.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = neon_matrix_impl.h; sourceTree = "<group>"; };
 		A08C379014144FDE004538FD /* neon_matrix_impl.c */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.c; path = neon_matrix_impl.c; sourceTree = "<group>"; };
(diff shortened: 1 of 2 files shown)


slembcke:
Recursive pausing was the compromise to fixing the (rather extreme) difficulty of using multiple schedulers correctly.

You patch is definitely not a safe thing to do. In your snippet above, self.pauseLayer.paused would already be NO. By removing the if statement, RecursivelyIncrementPausedAncestors() is being called even though the paused value didn't change. When you set self.paused = NO later, you'll have nodes that have negative _pausedAncestors values which is bad. Those nodes will be considered paused in very poorly defined ways. I can't merge this as is.

Instead of pausing the entire scene and unpausing certain parts, can you reorganize the parts you want to pause into a common sub-node?

iMobCoding:
Ok I supposed removing those lines was bad, but didn't have time to check what's happening...

Anyway, I already wrote that sub-node scene composition doesn't solve the issues with scheduled stuff in main scene. For example, lets say there is scheduled selector in main scene to fire in 1 second. In the meantime you press pause button. If you pause main scene, scheduler will be paused, but then you can't see animated pause layer. On the other hand, if you pause sub-node with the content, you can have animated pause layer, but main scene will fire a scheduler behind...

So I think there is a need for .paused property to behave differently. So if you pause root node, all children should be paused too. But you should be able to unpause some children branches if you need to:

```
scene
  |- node1
  |- node2
    |- node2a
    |- node2b
  |- node3
    |- node3a
    |- node3b
```

```
scene.paused = YES;
scene.node2.paused = NO;
```

Should make everything in the scene paused except node2, node2a & node2b

Best

slembcke:
While a pausing a node pauses the timers and actions of it's child nodes, it does not pause the nodes themselves. Doing that would prevent you from unpausing a parent node independently from child nodes. This is a problem that you would also probably run into, and it would be much harder to fix. You would have to track which nodes should stay paused when unpausing the parent.

Can you move the timers you are scheduling on the scene to a sub-node? A little reorganization would give you what you want without hacks.

```
scene
  |- pauseNodes (Add timers here, and pause this instead)
    |- node1
    |- node3
      |- node3a
      |- node3b
  |- node2
    |- node2a
    |- node2b
```

iMobCoding:
> While a pausing a node pauses the timers and actions of it's child nodes, it does not pause the nodes themselves. Doing that would prevent you from unpausing a parent node independently from child nodes. This is a problem that you would also probably run into, and it would be much harder to fix. You would have to track which nodes should stay paused when unpausing the parent.

Ok got it. But I actually thought exactly that: when you pause parent all children should be paused too. So when you unpause some child node, only it and its children unpause. When you again pause or unpause parent it propagates to all children again...

> Can you move the timers you are scheduling on the scene to a sub-node? A little reorganization would give you what you want without hacks.

Sure thing, but then I'd need to subclass pauseNode so I can fire timers? Or there is some other way I missed in v3? I just thought simple pause/unpause with parent to children propagation would be easier for everybody. Like it's already done for cascading colors/opacity...

Thanks

iMobCoding:
Ok, got it now. The whole purpose of why I asked for this **.paused** behavior was because my method for pausing everything didn't work:

```
-(void) pauseSchedulerAndActionsRecursiveFor:(CCNode *)parent  {
    [[[CCDirector sharedDirector] actionManager] pauseTarget:parent];
    [[[CCDirector sharedDirector] scheduler] setPaused:YES target:parent];

    for (CCNode *node in parent.children) {
        [self pauseSchedulerAndActionsRecursiveFor:node];
    }
}
```

And it didn't work because I triggered this bug:
https://github.com/cocos2d/cocos2d-iphone/issues/660

So, as soon as I added this:

```
    [[[CCDirector sharedDirector] scheduler] setPaused:NO target:parent];
    [[[CCDirector sharedDirector] scheduler] setPaused:YES target:parent];

```

Everything started working as expected. I am talking about scene's scheduled selectors which didn't want to pause, because each new scheduler is set as default to be paused.

ghost:
I think the pausing functionality needs more flexibility. When you pause a node, it recursively pauses all of its children nodes. 

Having a 'GameScene; say, that is the main node managing the game world (that is, it is the class that does most of the work relating to the game loop) and say a PausedHudNode (that will a child of the 'GameScene')  makes things iffy.

If I have a method to pause the game loop when the user hits the pause button, if I call self.paused = YES then the PausedHudLayer will never be displayed because the actions that display it will be paused (since its a child of the GameScene).

If I have my own method to pause all the children to avoid this problem, say something like:

> -(void)pauseMainGame
> {
>    myMainSprite.paused = YES;
>    enemySprite.paused = YES;
>   //ect....ect....
> 
>    [self showPausedGameLayerAnimated:YES];
> 
> }

Now the PausedHudNode will display (because the GameScene is not paused), but the GameScene's update method and scheduled methods will continue to run, further advancing the game loop even though nothing is happening because the user isn't playing. On v2 you just called pauseSchedulerAndActions on self to stop that. 

I don't think moving the game loop code to an empty node just to be able to pause it is really desirable. 

Perhaps, there should be a BOOL property to set on a node to block it from pausing with it's parent.

Birkemose:
It is all a matter of building your hierarchy right.

In your case, simply add an extra pause node as ancestor to everything you need to pause in your game scene, and add the rest to another node.The overhead of adding an extra control node is non existent.

ghost:
Yeah, I ended up doing that. I wrapped the 'GameLayer' class in a CCScene, (Cocos2d v2 style) and added the PauseHudNode as a child of the scene, so pausing the GameLayer won't effect the PausedHudNode. 

Def. the easiest way to handle this.

PR #677, opened 2014-04-03
FIX rsync error: some files could not be transferred

losingle:
install.sh is such a line of code， but LICENSE_Chipmunk.txt not found
copy_files "LICENSE_Chipmunk.txt" "$LIBS_DIR"


slembcke:
Whoops. I didn't realize some of the license files were referenced from install script.

I had to fix the issue in the older branches too so I did it by hand.

Issue #680, opened 2014-04-06
scale bug when build and run on first time on the retina display.


krazyeom:
![screen shot 2014-04-06 at 5 51 19 pm](https://cloud.githubusercontent.com/assets/442477/2624852/a5ba9194-bd68-11e3-863f-9af2099c149b.png)

there is something wrong when build and run on first time.

Issue #681, opened 2014-04-06
keyboard support for Mac


krazyeom:
Is there any plan to support a keyboard for mac?

krazyeom:
I've already patched to support a keyboard for mac. I am not sure it had some a problem. but it works.

Issue #632, opened 2014-03-06
2x scale on not full screen mode in Mac-retina graphics


krazyeom:
there is some bug on v3 RC 3. if is was not full screen mode, sprite is 2x scale.

slembcke:
Seems like this was taken care of in #633

PR #682, opened 2014-04-06
#issue-680 and #issue-681

krazyeom:
fixed scale bug #issue-680
# issue-681 support a keyboard for mac

### DIFF
--- a/cocos2d/CCDirector.m
+++ b/cocos2d/CCDirector.m
@@ -343,7 +343,7 @@ static CCDirector *_sharedDirector = nil;
 #ifdef __CC_PLATFORM_IOS
 		CGFloat scale = __view.layer.contentsScale ?: 1.0;
 #else
-		//self.view.wantsBestResolutionOpenGLSurface = YES;
+		self.view.wantsBestResolutionOpenGLSurface = YES;
 		CGFloat scale = self.view.window.backingScaleFactor;
 #endif
 		
--- a/cocos2d/CCResponder.h
+++ b/cocos2d/CCResponder.h
@@ -203,6 +203,20 @@
  */
 - (void)scrollWheel:(NSEvent *)theEvent;
 
+/**
+ *  Called whan a key down.
+ *
+ *  @param theEvent    Current event information.
+ */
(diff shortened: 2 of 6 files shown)


slembcke:
I'll close #680 and #681 now since they are handled here.

@Birkemose any thoughts on the CCResponder changes?

@vlidholt It looks like you commented out the retina line so that one would have to enable retina explicitly. I vaguely remember discussing that a couple months ago. It seems related to #632.

andykorth:
Issue 680, the scale bug seems to be fixed in develop-v4 basically as described here.

I'm going to take a look at Mac keyboard support using this commit. Thanks!

LearnCocos2D:
What's the status of #681 keyboard support for Mac? Currently documenting input events, just want to know if I should delay documenting keyboard events if it gets integrated in CCResponder soon.

Let me know if there are any obvious changes that need to be applied to the pull request, maybe I'll try documenting it as far as I can without actually being able to test the code.

Am I correct assuming that NSEvent will have to be converted to a CCKeyEvent object?

andykorth:
Basic Mac input support was added in #1191 and #1174. These two commits were worked into #1174, (since they were merged quite a while after the original contribution, there were some conflicts cleaned up in those branches). Github didn't really seem to pick up on the cherry-picking anywhere I guess.

So the status is that basic keyboard support works. There's support for nodes receiving keyboard and mouse up/down events. There's currently no way to check something like `iskeyDown:` in your update loop. For now that can be pretty easily implemented in game code.  It would be great to add that by v4, but I'm not sure where it falls priority-wise.

I don't think the stuff that's currently in there is likely to move or change, so it can be documented. The only thing that might change is the 'other' mouse button: https://github.com/cocos2d/cocos2d-swift/issues/1234

LearnCocos2D:
Ah thanks, I could not find the slightest reference to "key" events in CCResponder or CCResponderManager in v3.4.1. But it's there in develop (v3.4.3). I didn't expect this to be a very recent addition.

andykorth:
Oh sorry, yeah, the changes we did are in develop for v4.

There was no mac keyboard support in any v3.

PR #688, opened 2014-04-09
Update CCTiledMap.m

smklinger:
This will ignore empty tilemap layers, if the desired functionality is to allow empty layers to be added, CCTiledMapLayer will have to be changed to allow for 'nil' tilesets.

CCTiledMapLayer:setupTiles doesn't check for a 'nil' tileset before trying to change its properties.


smklinger:
Replaced with pull request #694

PR #691, opened 2014-04-10
 Fixing memory leak

iSevenDays:
(CGColorRef) CGColor { CGFloat components[4] = {(CGFloat)_r, (CGFloat)_g, (CGFloat)_b, (CGFloat)_a}; return CGColorCreate(CGColorSpaceCreateDeviceRGB(), components); }
CGColorSpaceCreateDeviceRGB() returns a Core Foundation object with +1 retain count and this object is not referenced later and has a retain count of +1.

### DIFF
--- a/cocos2d/Support/CCColor.m
+++ b/cocos2d/Support/CCColor.m
@@ -170,10 +170,14 @@
 }
 #endif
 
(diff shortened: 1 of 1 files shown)


slembcke:
Thanks.

PR #692, opened 2014-04-10
Physics additions to cocos2d for animations.

jtwigg:
Lots of changes:
Static bodies that animate turn into kinematic bodies. Velocities and angular velocities are injected.
Parents of nodes with physics can have animations on them. 
These animations will run on a fixed timestep


jtwigg:
updated with fixed unit tests.

slembcke:
Works great and is relatively straightforward.

The KVO idea looks like it's a lot cleaner than what I had suggested. Keeps a lot of the code out of CCNode. If the performance is okay to use it on every physics enabled node, I'm wondering if we could use that to clean up even more of the physics code.

Issue #693, opened 2014-04-10
[TMX ERROR] an empty layer, all display blank 


EarthAlchemist:
When make a tmx file by tiled map editor, I create an empty layer to contain a player ccSprite in runtime, but when I add the tmx to scence, all display blank...

when I fill some tile to the empty layer, the tmx is display in the scence.

In the code, CCTileMap.m line 87, the empty layer will make the contentsize.width and height to +INF, and this is error...

smklinger:
I traced the problem to CCTileMapLayer.m line 120. If the layer doesn't have a tileset, pixelsToPoints is 1/0 = INF which causes the layer's contentSize to be INF, INF

I've submitted pull request #694

Issue #687, opened 2014-04-09
CCTiledMap doesn't parse new CCTiledMapLayers after encountering one with zero tiles


smklinger:
in CCTiledMap's private "parseLayer" function after getting CCTiledMapTilesetInfo, it should just return nil if tileset is nil.

smklinger:
Should empty layers be ignored? or added? I'm not sure what the desired functionality is.

EarthAlchemist:
In the code, CCTileMap.m line 87, the empty layer will make the contentsize.width and height to +INF, and this is error...

smklinger:
I've submitted pull request #694

PR #694, opened 2014-04-10
CCTiledMapLayer.m issues #687 & #693

smklinger:
Fixes possible divide by zero. If the layer has no tileset, the tex is nil and causes pixelsToPoints to be INF. This causes the contentSize to be INF, INF and prevents any further layers from being parsed.


EarthAlchemist:
OK!

slembcke:
Thanks.

PR #661, opened 2014-03-27
Cocoapods

dylan:
Hello everyone. I put together a podspec for Cocos2d-iPhone 3.0. So far I've been able to run the sample project by starting with a template, removing the libs from the project, removing the objective chimpunk target and a making a podfile resembling this: https://gist.github.com/dylan/9797584 and running `pod install` .

Hopefully this can act as a starting point for making cocos2d/chipmunk Cocoapods compatible.

Please try it out! :smile:

PR #696, opened 2014-04-11
Added updated podspec

dylan:
Picking up where I left off at #661.


slembcke:
Part of the reason why this has been left unmerged for so long is because I don't think any of the main Cocos contributors use CocoaPods. Finally reading up on this now...

dylan:
:+1: It's extremely popular with the general iOS dev community.

qichunren:
Many people usually use SpriteBuilder to generate cocos2d-iphone Xcode project, and I think this is more effective than using CocoaPod.

richardgroves:
Seriously? Do you have any data on that?

I know that SpriteBuilder is being heavily pushed these days - but (also
without data) I'm pretty sure the majority of people start their projects
from XCode.
And given the issues related to updating an existing project for even minor
version updates I suspect a solid, supported, CocoaPod release of the
library would be most welcome.

My main issue is that I'm not sure how CocoaPod handles a source-available
project: can it be made to add all the library source in or does it only
work with a .a version of projects - it is _so_ helpful to have the source
actually in the project when trying to figure out how things work.

On Tue, Jul 1, 2014 at 8:13 AM, Qichunren notifications@github.com wrote:

> Many people usually use SpriteBuilder to generate cocos2d-iphone Xcode
> project, and I think this is more effective than using CocoaPod.
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/cocos2d/cocos2d-iphone/pull/696#issuecomment-47624029
> .

## 

Richard Groves
Technical Director | NoodlFroot

M: +44 (0)7962 054 163

www.noodlfroot.com

NoodlFroot Limited is a company registered in England and Wales with
registration number 07245048.
Registered office: 145-157 St John Street, London, EC1V 4PY.

dylan:
This podspec uses the library source.

dylan:
Still works, I used it to get a blank Swift project using cocos2d...

jshier:
An up-to-date and maintained podspec would be most welcome. I'm not sure who owns the Cocos2D spec at the moment but it's at 3.1.0-alpha, which isn't ideal.

lysanntranvouez:
:+1: 

In case anybody is interested: I made a podspec for 3.4.3-rc0: https://gist.github.com/lysannschlegel/26373d2f45d18527c298

I would really love to see an officially maintained podspec, though. (Partly because I had to hardcode all the file names because patterns wouldn't do. Sometimes I had to add Android files, sometimes not.....)

I'm a developer and therefore have some preferences about tools I want to use. So I wasn't thrilled to see that I was supposed to download an extra application just for getting the newest Cocos2D release.
Especially given that after upgrading from an old unsupported version, integration into my existing project was ... not straight-forward at all. I had to create a new empty project, just to see what flags to set... :frowning:

Cocoapods is the de-facto standard in the iOS community. And even if people don't want to use it (they're developers, they have preferences :wink:), I'm pretty sure they'd rather integrate manually, with some detailed instructions (those instructions wouldn't even be too complicated), instead of being told to use an external application.

Source: I think so :wink:

Issue #663, opened 2014-03-27
CCbutton inside a CCScrollview steals touch and breaks all touch events


fgeorg:
Here's a video showing the issue: https://www.dropbox.com/s/50nrn79ggcyozwm/ccbutton_bug.mov

it looks like it gets in a messed up state when the touch ended command happens outside of the button and there's no touch moved event. (this is only possible in a ccscrollview)

mattcn:
I have got the similar issue. 

Sometimes, when scrolling the table/scroll view. the touch control outside the scrollview can not be detected.

TimPelgrim:
I have this issue in multiple projects. There is a tempory fix suggested here: http://www.cocos2d-iphone.org/forums/topic/v3-ccbutton-in-ccscrollview-cause-untouchable/#post-416672 but this breaks the interaction with a scrollview in that it can't be stopped while animating.

NickyWeber:
Bugfix on the way: https://github.com/cocos2d/cocos2d-iphone/pull/701

slembcke:
#701 is merged into branch v3.0. Thanks everyone.

ghost:
I'm getting this bug right now, not with a button but with a node subclass with userInteractionEnabled. 

When this node subclass is put inside the CCScrollView, and I press the node I get the touchBegan...but if I move my finger outside the destination (on the same touch), I get touchMoved:withEvent but not touchEnded:withEvent:

Scroll view just steals the touch.

I see the scroll view uses a UIPanGestureRecognizer object instead of doing touch handling completely manually at the UITouch level. Not sure if maybe that is the reason...UIKit grabbing the same touch my node is working on? 

I think a great thing to add to Cocos2d would be our own gesture recognizer objects instead of always having to reimplement lots of the same touch-handling code in lots of different places. Would be nice to consolidate gestures into objects. While we can add UIGestureRecognizers to the director's view, it would be sweet if we had our own gesture recognizers that we could just add to node's directly like UIKit.

NickyWeber:
@AT2014 normally if you move your finger outside the screen you should get a touchCancelled event like in UIKit.

Please post a separate feature suggestion, this issue is closed and won't get noticed.

PR #701, opened 2014-04-15
Bugfix: CCbutton inside a CCScrollview steals touch and breaks all touch events #663 

NickyWeber:
Scrollview will stop panning if touched and won't steal touch anymore.
CCTapDownGestureRecognizer touch began phase implemented with very old iOS interface.


slembcke:
I have never noticed that bug until I tried to replicate it. It's almost unbelievable that I never triggered it. Apparently it's been there for months. O_o

PR #706, opened 2014-04-16
Bugfix/#705 set blend func not showing immediately

NickyWeber:
RenderState is not invalidated when blendFunc changes. 
Solution: reuse blendMode setter, instead of setting the value directly. That will invalidate the renderState if the blendMode changes.


NickyWeber:
Damn, wrong branch, has to be develop.

Issue #705, opened 2014-04-16
Setting CCNode's blendFunc won't show...


NickyWeber:
until the renderState is invalidated by some other action.

slembcke:
Whoops. I should have used the property setter. Fixed.

PR #707, opened 2014-04-16
Bugfix: Setting CCNode's blendFunc won't show #705 

NickyWeber:
RenderState is not invalidated when blendFunc changes.
Solution: reuse blendMode setter, instead of setting the value directly. That will invalidate the renderState if the blendMode changes.


slembcke:
Err, whoops. I guess I shouldn't have read the issues in order. Already pushed a fix for this after reading the other issue. Sorry about that.

NickyWeber:
All good, fix is in, that's rad.

PR #711, opened 2014-04-24
fixed buildResponderList: in CCResponderManager

kalo:
previously _buildResponderList:_ overlooked adding the current node to the responder list if the node _only_ contained children with a negative z-index while having interactions enabled.

flagging the node for possible addition, then eventually adding it should patch this.

and some minimisations.

### DIFF
--- a/cocos2d/CCResponderManager.m
+++ b/cocos2d/CCResponderManager.m
@@ -103,31 +103,29 @@
     _dirty = NO;
 }
 
--( void )buildResponderList:(CCNode *)node
+- (void)buildResponderList:(CCNode *)node
 {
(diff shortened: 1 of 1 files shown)


slembcke:
Had to read it several times to see what the bug was, or what the change was doing. O_o I made a similar change to [CCNode visit:...] in 3.1 to work around ARC performance issues.

kalo:
Apologies for my flavoured wording. 

It actually took me a little while to understand that the responder was actually trying to do the same thing as `visit` and not something else completely. 

I've made the same change in my node's `visit` locally as well, but never thought to push it because it wasn't broken to begin with.

I suppose it also makes it all look cleaner

PR #712, opened 2014-04-25
Update CCParallaxNode.m

smklinger:
_parallaxArray is an array of CGPointObjects, not an array of nodes. We should search for the CGPointObject that points to the CCNode to be removed, and remove it.

### DIFF
--- a/cocos2d/CCParallaxNode.m
+++ b/cocos2d/CCParallaxNode.m
@@ -95,7 +95,27 @@
 
 -(void) removeChild:(CCNode*)node cleanup:(BOOL)cleanup
 {
-	[_parallaxArray removeObject:node];
+
+	CGPointObject *objectToRemove = nil;
+    
+	for (CGPointObject *point in _parallaxArray) {
+	    
+		if(point.child == node){
+
+			objectToRemove = point;
+            
+			break;
+            
+		}
+    
+	}
+    
+	if(objectToRemove != nil){
+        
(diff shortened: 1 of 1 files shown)


smklinger:
Yeah, that does look better. What do you think of this?

```
    for(NSUInteger i = 0; i < _parallaxArray.count; i++){

        if(((CGPointObject *)[_parallaxArray objectAtIndex:i]).child == node){

            [_parallaxArray removeObjectAtIndex:i];

            break;

        }

    }
```

kalo:
at some point in the past i definitely would have opted to do a **for loop + objectAtIndex:** and it would lead into something like that. 

however, as a more modern solution,  **fast enumeration** will likely hold up better.

the _for loop objectAtIndex_ is only slightly faster than _fast enumeration_ when there are only a hand full of objects in the array. fast enumeration is consistently faster going into dozens to hundreds of objects.

my solution for objectAtIndex: would probably be something like this

``` objc
[super removeChild:node cleanup:cleanup];

for (NSUInteger i = 0; i < _parallaxArray.count; i++){
   if (((CGPointObject *)[_parallaxArray objectAtIndex:i]).child == node){
      [_parallaxArray removeObjectAtIndex:i];
      return;
   }
}
```

i put the `super` call at the beginning so we can just return from the function as soon as the object is removed from the parallax array. i'm quite old fashioned when it comes to `break`ing.

in either case, faster enumeration is probably the best solution.

PR #713, opened 2014-04-25
fixed small typo in comment in default cocos2d template

ghost:
I know this 'fix' is pretty useless and obvious for everyone but few lines above comment which I changed we have:

``` obj-c
self.userInteractionEnabled = YES;
```

so in my opinion we should keep the names case sensitive (like in other cases).

### DIFF
--- a/templates/cocos2d iOS.xctemplate/Classes/HelloWorldScene.m
+++ b/templates/cocos2d iOS.xctemplate/Classes/HelloWorldScene.m
@@ -80,7 +80,7 @@
     [super onEnter];
     
     // In pre-v3, touch enable and scheduleUpdate was called here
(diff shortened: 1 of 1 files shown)


templates/cocos2d iOS.xctemplate/Classes/HelloWorldScene.m
```diff
@@ -80,7 +80,7 @@ - (void)onEnter
     [super onEnter];
     
     // In pre-v3, touch enable and scheduleUpdate was called here
-    // In v3, touch is enabled by setting userInterActionEnabled for the individual nodes
+    // In v3, touch is enabled by setting userInteractionEnabled for the individual nodes
```
ghost, line 83: This comment refers to the line 40 in this file when we are using:

``` obj-c
self.userInteractionEnabled = YES;
```

PR #714, opened 2014-04-25
Joints support scaling.

jtwigg:


slembcke:
Sorry for the delay.

I'd prefer if it was consistent with how CCShapes worked (updating the scale from the original parameters in willAddToPhysicsNode: instead of storing an explicit scale), but otherwise don't see any issues with it. Can we add some unit tests for this so that it's possible to update the implementation in the future without breaking it?

A potential problem is that the scale is only sampled as the scale of the first body that the joint connects to. If the joint is connected between two bodies that don't share the same hierarchy it might do something really unexpected. It doesn't seem like there is a good one-size-fits-all solution, so this will have to do. For a point of reference, Unity doesn't seem to apply scaling to joints at all.

PR #717, opened 2014-04-26
Update CCParallaxNode.m

smklinger:
_parallaxArray is an array of CGPointObjects, not an array of nodes. We should search for the CGPointObject that points to the CCNode to be removed, and remove it.


smklinger:
Replacing #712 with this. Thanks kalo!

kalo:
excellent!   `=)`

slembcke:
Hmm. Just merged this, though it got marked as a regular commit instead of a merge so GitHub didn't pick up on it. Not sure how I did that.

Thanks all, sorry for the review delay.

Issue #698, opened 2014-04-13
setSprite: in CCProgressNode doesn't work!


FuriouZz:
When I set a new sprite, the CCProgressNode sprite display nothing.

My solution is to add `[self updateProgress];` at the end of the second condition of the `setSprite:` like in `setType:`.

kalo:
i just made a pull for this.

`setReverseDirection:` also created the same problem.

FuriouZz:
Thanks a lot for your reaction. I'm waiting your pull.

slembcke:
@kalo's pull request was merged (finally). Thanks everyone!

PR #719, opened 2014-04-27
properly update vertices for CCProgressNode

kalo:
CCProgressNode only properly updated visually when the progress percentage or type was changed and ignored doing the same if the sprite (#698) or direction changed.

flags were added with intention for batching the differences to avoid redundant calculations and will update if the node is visible and dirty before drawing rather than when set.

directly set some ivars in `init` to avoid some degree of redundancy.
`return`s replaced `break`s in the switch because it will be microscopically faster.
and some other minor changes.

on a side note, there are a few other possible improvements to be made, but those are for another day.


slembcke:
Sorry it took a while to get this reviewed.

PR #720, opened 2014-04-27
Add CCRenderer support for Cocos3D rendering.

billhollings:
Make `CCRenderCommand` protocol public. Rename `CCRenderCommand` invoke: method to
`invokeOnRenderer:` for more clarity as public protocol applicable to any object.
Add `CCRenderer enqueueRenderCommand:` method.
Set `COCOS2D_VERSION` to `0x00030100` so Cocos3D can detect new rendering.

Within Cocos3D, I've created implementations of `CCRenderCommand` to render Cocos3D components.  Exposing the `CCRenderCommand` as public, and adding the basic `enqueueRenderCommand:` supports this, and will allow any Cocos2D developer to create customized commands for their own rendering needs.

Finally, in order to provide support for multiple versions of Cocos2D, Cocos3D uses the `COCOS2D_VERSION` setting to detect the version. To allow such support, we should get in the habit of making this the first thing we change as we open a branch for an upcoming version.

### DIFF
--- a/cocos2d/CCRenderer.h
+++ b/cocos2d/CCRenderer.h
@@ -25,7 +25,7 @@
 #import "ccTypes.h"
 
 
-@class CCTexture;
+@class CCTexture, CCRenderer;
 
 /// Standard interleaved vertex format for Cocos2D.
 typedef struct CCVertex {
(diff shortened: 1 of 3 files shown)


slembcke:
I moved the protocol and method definition to the CCRendere_Private.h header for now. In the fairly near future, there will need to be some modifications  to mark if a particular command is thread safe or not.

[CCRenderer enqueueBlock:] should work for most things that a custom render command class would except for mutability. (Like how CCRendererDrawCommand is updated for two sequential draw commands)

Issue #723, opened 2014-04-29
CCRenderTexture inverts near / far plane in projection matrix.


hakonber:
In Cocos2D v3.0, the current CCRenderTexture implementation inverts the near and far planes in the projection matrix that's setup in -(void)begin. This breaks apps that rely on depth testing for correct rendering.

Relevant code:

``` objc
-(void)begin
{
    ...
    kmMat4 orthoMatrix;
    kmMat4OrthographicProjection(&orthoMatrix, (float)-1.0 / widthRatio,  (float)1.0 / widthRatio,
                                 (float)-1.0 / heightRatio, (float)1.0 / heightRatio, -1,1 );
    kmGLMultMatrix(&orthoMatrix);
    ...
}
```

You can observe that the near / far plane gets inverted by stepping into the kmGLMultMatrix() call and checking the matrix before and after the multiplication.

The fix is easy, swap the -1, 1 parameters in the kmMat4OrthographicProjection() call.

PR #724, opened 2014-04-29
Fix for inverted near / far planes in CCRenderTexture projection matrix. 

hakonber:
See issue #723


slembcke:
I double checked that the issue is fixed in 3.1 as well. Apparently I did it correctly when rewriting it. \o/

Issue #728, opened 2014-05-01
CCLayoutBox with vertical direction: sort order is inverse of expected order


LearnCocos2D:
If a layout box has 4 children, then the child nodes are displayed bottom up, ie 4th child node is at the top and 1st child node at the bottom. 

Best seen in SpriteBuilder:
![screen shot 2014-05-01 at 17 14 31](https://cloud.githubusercontent.com/assets/214599/2853368/7536302a-d143-11e3-9543-a1fd2a920779.png)

This happens only when direction is set to vertical, in horizontal mode the order of nodes is logical (quit is to the left, resume to the right).

kalo:
this should be a simple fix. 

accumulate the maximum height while searching for the max width, and subtract the rolling height from the max height as the nodes are organised.

vlidholt:
This is not a bug, but I guess it can be a matter of discussion for how to do this. Bottom up, like the coordinate system – or top down like the UI often is layout.

This is, however, not an easy fix as changing this will break any ccb files in SpriteBuilder.

kalo:
I didn't think of the effects of this in spritebuilder because I have yet to use it. 

I suppose a third layout direction for top down could be added eventually

vlidholt:
Yeah, that could be a better option, or a property that defines the direction.

iMobCoding:
I gave a much simpler fix some time ago:
https://github.com/cocos2d/cocos2d-iphone/pull/563

slembcke:
Yeah... I remembered that issue but hadn't gotten around to checking out what happened to it. Is this a regression or did somebody change their mind?

iMobCoding:
Yes it was committed then reversed for the same reason Vik posted here - breaking things in Sprite Builder... I agree, adding a property for 3rd layout direction would be the best. And making it the default for vertical direction

ryanmarfone:
Top down and vert would be nice defaults

PR #730, opened 2014-05-02
CCLayoutBox vertical layout from top down

kalo:
fix for issue #728.

this should layout nodes from the top down, rather than bottom up.


vlidholt:
This is not a bug, but I guess it can be a matter of discussion for how to do this. Bottom up, like the coordinate system – or top down like the UI often is layout.

This is, however, not an easy fix as changing this will break any ccb files in SpriteBuilder.

PR #731, opened 2014-05-02
CCEffects

osinoleg:
Added CCEffects - not really ready for consumption yet.

After adding multi pass rendering support, I had to break a lot of functionality for CCEffects - here is a list of things that need to be fixed.

Main issues:
- CCEffectNodes that use mutli pass effects will not auto draw correctly (as in, you can't add the effectnode correctly to the scene, although you can treat it like a ccrendertexture and just call [node visit], then use the effectnodes texture).
- CCEffects with mutli pass rendering don't stack - work in progress.
- CCEffectNodes output texture does not blend correctly with the scene.

Minor issues:
- I stubbed out the ability to set CCEffects on CCSprites for now, waiting to resolve the above issues first.
- Allow more than one CCShader to be used per CCEffect, this will help with optimization and avoid a crazy uber/branching shader.


vlidholt:
Some comments:

All interface additions needs to be hidden behind #ifdef:s. How about CC_ENABLE_EXPERIMENTAL_EFFECTS in ccConfig.h

I think the effects should have their own test case, not bundled with the renderer. Each effect should have their own sub-test.

osinoleg:
I 100% agree, will do that now!

slembcke:
Seems like everything is still on track. Still some things to figure out relating to (pre-multiplied) alpha

You shouldn't need to call [CCRenderer flush] explicitly from [CCEffectNode draw:transform:]. From skimming the code, I couldn't figure out why it was needed exactly. Does it synchronize the rendering with changes to the effect node's sprite's render state changes? Flush was originally intended to be called once per frame. Not a priority now, but it will need to change at some point for threaded rendering where flushing the renderer will be moved to a different thread.

PR #758, opened 2014-05-15
convertContentSize(To/From)Points Yields 0 on Normalized

josejulio:
When using convertContentSize(To/From)Points and you have either CCSizeUnitNormalized, CCSizeUnitInsetPoints or CCSizeUnitInsetUIPoints on your top node, you will get 0 for that one, because the top node has no parent. (nil.something == 0).

### DIFF
--- a/cocos2d/CCNode.m
+++ b/cocos2d/CCNode.m
@@ -418,15 +418,27 @@ TransformPointAsVector(CGPoint p, CGAffineTransform t)
     }
     else if (widthUnit == CCSizeUnitNormalized)
     {
-        size.width = contentSize.width * _parent.contentSizeInPoints.width;
+        if (_parent) {
+            size.width = contentSize.width * _parent.contentSizeInPoints.width;
+        } else {
+            size.width = contentSize.width * director.viewSize.width;
+        }
     }
     else if (widthUnit == CCSizeUnitInsetPoints)
     {
-        size.width = _parent.contentSizeInPoints.width - contentSize.width;
+        if (_parent) {
+            size.width = _parent.contentSizeInPoints.width - contentSize.width;
+        } else {
+            size.width = director.viewSize.width - contentSize.width;
+        }
     }
     else if (widthUnit == CCSizeUnitInsetUIPoints)
     {
-        size.width = _parent.contentSizeInPoints.width - contentSize.width * director.UIScaleFactor;
+        if (_parent) {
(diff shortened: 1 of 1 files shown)


slembcke:
Perhaps it should log a warning instead, but returning the size relative to the director doesn't seem correct to me. If the size is relative to the parent and the node doesn't have a parent, then the size is undefined. It seems like it might be even more confusing to have it return a screen relative size instead of 0.

@vlidholt Thoughts on this one?

josejulio:
I was thinking that if a node does not have a parent it might be the "top node" and its _parent_ might be the screen.
It happened to me that i was trying to get the size on points of a node, but always returned 0 and had no idea what was happening, turns out that the top node had a relative size (i used SpriteBuilder to get this conf)

vlidholt:
I think that returning 0 is the correct value to return, but I can see the case where this confuses users. Not sure what the best way to handle this is, better explanation in the docs or if it should log a message.

josejulio:
I understand, just a though, maybe there could be a "Screen" node to by-pass this issue?
i.e. my top node can be set to % but it will have a parent with the exact dimensiones of the screen, thus nodes attached to the screen will have a good value and those who aren't will return 0.

Ben-G:
+1 for "Screen" node. I think it's a nice compromise between usability and a "correct" implementation.

danielhaaser:
@josejulio I think that's a good idea.  Unfortunately, returning (0, 0), may be strictly correct but it is both counter-intuitive and adds extra cognitive load to what should be easy positioning code.

richardgroves:
Could it/should it at least throw a debug log message when a parentless
node with normalised position/content size gets asked to return it? That
would help people debug the issue without having to make any fundamental
changes.

On Thu, Jul 10, 2014 at 5:15 PM, djh- notifications@github.com wrote:

> @josejulio https://github.com/josejulio I think that's a good idea.
> Unfortunately, returning (0, 0), may be strictly correct but it is both
> counter-intuitive and adds extra cognitive load to what should be easy
> positioning code.
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/cocos2d/cocos2d-iphone/pull/758#issuecomment-48627184
> .

## 

Richard Groves
Technical Director | NoodlFroot

M: +44 (0)7962 054 163

www.noodlfroot.com

NoodlFroot Limited is a company registered in England and Wales with
registration number 07245048.
Registered office: 145-157 St John Street, London, EC1V 4PY.

PR #764, opened 2014-05-19
CCLayoutBox in vertical order was sorted in reverse order.

NickyWeber:
Fixes: https://github.com/spritebuilder/SpriteBuilder/issues/280


vlidholt:
Merging this pull request will break compatibility with older versions. How the layout box is sorted top to bottom or the way the the coordinate system works is a matter of preference.

NickyWeber:
Alright, np, I'll link the bug to here.
Btw, it might be a good idea to decouple the z-ordering from the ordering and figure another way, like drag and drop to sort a layout.

PR #773, opened 2014-05-22
CCNodeGradientRadialColor class: draw radial gradient with GL_TRIANGLE_FAN 

chunkyguy:
CCNodeGradientRadial is a CCNode that can be used to display radial gradient

 It isn't a subclass of CCNodeColor or CCNodeGradient because:
1. CCNodeColor only has 4 vertex available whereas for a gradient effect we need more vertices.
2. CCNodeColor uses GL_TRIANGLE_STRIP for rendering but for radial gradient we're using GL_TRIANGLE_FAN

Still the CCNodeGradientRadial's interface is kept as close to CCNodeGradient so that the gradient effect could be switched with minimum efforts.

Here's a screenshot from my game in development:
img_0680

I searched the internet for a way to draw radial backgrounds, but most of the people were suggesting using a texture or implementing using a fragment shader. I think with GL_TRIANGLE_FAN we can get a descent enough result with less performance overhead compared to the other two suggestions.

I've also added a test case and here's the result:

screenshot 2014-05-22 07 59 57

I'm new to the cocos2d and I'm still at level 0 with git. The cocos2d repository is already so complicated for me, sorry for anything stupid that I might have already done. Please let me know how it works out. 

Thanks for all your great work, I would love to get more involved in the future with the development :)
- Sid.

### DIFF
--- a/cocos2d-ui-tests/tests/ColorTest.m
+++ b/cocos2d-ui-tests/tests/ColorTest.m
@@ -167,15 +167,14 @@
 -(void) setupCCNodeGradientRadialColorTest
 {
 	CGSize s = [[CCDirector sharedDirector] viewSize];
-    
+
 	CCNodeGradientRadial *colorNode = [CCNodeGradientRadial nodeWithColor:[CCColor blueColor]
-                                                     fadingTo:[CCColor redColor]
-                                                  alongVector:ccp(1, 1)];
+                                                     fadingTo:[CCColor greenColor]];
 	colorNode.contentSize = CGSizeMake(200, 200);
 	colorNode.position = ccp( s.width/2.0f - 100, s.height/2.0f - 100);
 	[self.contentNode addChild:colorNode];
 	
-	self.subTitle = @"Blue inner, red outer";
+	self.subTitle = @"Blue inner, green outer";
 }
 
 -(void) unfinishedsetupBMFontColorCascadeTest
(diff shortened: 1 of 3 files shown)


slembcke:
I've been thinking about this for a bit. There are a few things to consider. First is that the custom rendering API changed in 3.1 so it would need to be rewritten significantly to support it going forward. The second is that it involves a _lot_ of code to avoid using a texture. Almost as much complexity as the sprite rendering.

A 4kb 256x256 RGBA texture would provide at least as much quality and performance as a triangle fan but without all the extra geometry, and geometry does get a bit more expensive in 3.1 since it optimizes for batching instead. In 3.1, a 1kb luminance texture + shader would be just as performant. A single sqrt() in a shader might be just as fast as the texture lookup too. Either option would be a fraction as much code and more 3.1 friendly.

Birkemose:
I think stuff like this in the future could be added to the cocos2d-iphone-ext repo.
If people then like a specific component, and it fits into the official API, we can move it to the official repo at next release.

Den 29 May 2014 kl. 18:15 skrev Scott Lembcke notifications@github.com:

chunkyguy:
I originally wanted to commit to the cocos2d-ext repo. I knew this class can run out of cocos2d core as an extension. But, I was unable to build the cocos2d-ext repo on my system. I'm not sure what the best place to ask these sort of questions, so I posted on the cocos2d forum but got no response, I don't know how active the extension repo is. 

http://forum.cocos2d-iphone.org/t/working-on-an-new-extension-getting-fatal-error-at-submodule-update/13487

Going forward I would love to add few more extensions that I'm working on right now, if it gets fixed. Those are mostly add-ons type components.

PR #782, opened 2014-05-23
CCLabelBMFont - _childForTag needs to be cleared when calling setFntFile...

jwarhurst:
....   This fixes an issue with bmfonts in spritebuilder displaying incorrectly.

Before Picture:

image


slembcke:
Ah. Good catch.

PR #784, opened 2014-05-23
Added ellipse recognition to the tmx-parser.

janwillms:
The CCTMXXMLParser now recognizes the ellipse-element `<ellipse/>` inside an object-group and adds an boolean (NSNumber, true) to the dictionary if the object uses the ellipse-tag.


slembcke:
Handy. I just noticed that was missing this morning while adding more tilemap tests.

PR #786, opened 2014-05-24
Remove faulty assertion tests in CCES2Renderer + support for rotations between landscape & portrait iOS device orientations.

billhollings:
Remove faulty assertion tests in CCES2Renderer, to permit zero as a valid framebuffer and renderbuffer ID under Android. Plus add support for iOS device rotation between landscape and portrait.

### DIFF
--- a/cocos2d/CCDirector.h
+++ b/cocos2d/CCDirector.h
@@ -230,9 +230,9 @@ and when to execute the Scenes.
 /**
  *  Changes the projection size.
  *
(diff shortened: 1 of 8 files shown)


slembcke:
Seems good to me. Somebody else was talking about supporting orientation changes (@Birkemose maybe?). I'm wondering if they have any input on this. I'll merge it tomorrow if there are no objections.

PR #790, opened 2014-05-27
Fixing documentation typos - CCNodeMultiplexer

richardgroves:
Fixing a few small documentation typos on CCNodeMultiplexer.

On a wider note should this class really be in CCNodeColor.h/m?

### DIFF
--- a/cocos2d/CCNodeColor.h
+++ b/cocos2d/CCNodeColor.h
@@ -211,7 +211,7 @@
  *  Features:
  *
  *  - It supports one or more children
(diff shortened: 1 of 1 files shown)


slembcke:
It is weird that it's in CCNodeColor.h... I've never actually used that class myself. The functionality it replaces (like 3 lines of code) is so trivial I'm not sure what it was intended for.

richardgroves:
Yep - probably a prime candidate for moving to a 'stuff that exists but we'd rather get rid of file'

PR #792, opened 2014-05-27
Rework Animation System lifecycle and API.

jtwigg:


cocos2d/CCNode.m
```diff
@@ -208,9 +208,8 @@ - (void)cleanup
 	[_children makeObjectsPerformSelector:@selector(cleanup)];
     
     // CCBAnimationManager Cleanup (Set by SpriteBuilder)
-    if([_userObject isKindOfClass:[CCBAnimationManager class]]) {
-        [_userObject performSelector:@selector(cleanup)];
-    }
+    [_animationManager performSelector:@selector(cleanup)];
```
slembcke, line 211: Why is this not [_animationManager cleanup]?

cocos2d/CCSprite9Slice.m
```diff
@@ -176,7 +176,11 @@ -(void)draw:(CCRenderer *)renderer transform:(const GLKMatrix4 *)transform
 - (float)margin
 {
     // if margins are not the same, a unified margin can nort be read
-    NSAssert(_marginLeft == _marginRight == _marginTop == _marginBottom, @"Margin can not be read. Do not know which margin to return");
+    NSAssert(*(unsigned int *)&_marginLeft |
```
slembcke, line 179: This isn't really 64 bit safe, what is the point of this?

edit: Derp. I thought maybe it was to avoid a floating point comparison warning, but the original was quite wrong. Shouldn't be casting to non-explicitly sized integers though. Changed here: f18678b50c7a4bfb5c4212bad64cb9b0a24d725d

PR #814, opened 2014-06-04
fix debug define & animation Divergence.

jtwigg:


slembcke:
`DEBUG` is used inconsistently. In most instances it's values must be 1, but in others it simply needs to be defined. That's half of the problem here.

Initially I thought we should always enforce the 0/1 value and not just the definition. I'd run into problems in the past because of it, but I think that was due to the build systems that were being used. We should update all of the instances to use `#ifdef DEBUG`. I'll take care of that next week when I merge this. (Unless you want to update the pull request in the mean time? That would be handy.)

There is also `COCOS2D_DEBUG`, but that is used specifically for controlling logging verbosity.

For comparison, both assert.h and NSException.h use a simple macro definition. If the macro is defined at all, disable assertions:

```
#ifdef NDEBUG
#define assert(e)   ((void)0)
#else
```

```
#if !defined(NS_BLOCK_ASSERTIONS)
// Enable assertions
#endif
```

I do really wish there was a standard macro for debugging though... Oh well. Maybe in Swift...

jtwigg:
Its a left over C++'y thing. Don't fight the waters. Just go limp.

jtwigg:
Cross reference: https://github.com/cocos2d/cocos2d-iphone/issues/818

PR #850, opened 2014-06-26
Fixes for high priority issues and other smaller issues.

cocojoe:
Investigating SB issues, turns out that an empty timeline will still always play the default timeline and then perform the sequenceComplete selector. (Not sure if it needs to but always has so leaving it alone).
CCActionCallFun retains this.

CCActionCallFunc set target node to weak,
Setup scheduler in CCAnimationManager only if added to a node (onEnter)

spritebuilder/SpriteBuilder#638
spritebuilder/SpriteBuilder#653
spritebuilder/SpriteBuilder#654


cocojoe:
@slembcke / @vlidholt  Any chance you can review please, ideally could do with pushing these changes out to next SB AppStore build.

vlidholt:
This update needs to go into the v3.1 branch and then merge into develop.

cocojoe:
Done #854

PR #854, opened 2014-06-27
Fixes for high priority issues and other smaller issues.

cocojoe:
Investigating SB issues, turns out that an empty timeline will still always play the default timeline and then perform the sequenceComplete selector. (Not sure if it needs to but always has so leaving it alone).
CCActionCallFun retains this.

CCActionCallFunc set target node to weak,
Setup scheduler in CCAnimationManager only if added to a node (onEnter)

spritebuilder/SpriteBuilder#638
spritebuilder/SpriteBuilder#653
spritebuilder/SpriteBuilder#654
spritebuilder/SpriteBuilder#658
spritebuilder/SpriteBuilder#659


jtwigg:
I had a read and it all seems correct, especially the opacity fix. 
It also passes the SB animation unit tests.
@slembcke  @vlidholt  I don't have permission to merge, please do so.

PR #864, opened 2014-07-02
Fix crash sequence completion

orestrnp:
<b>Additions</b>
FixCrash - when a sequence completes get the sequence from the list by filtering for its sequenceId instead of accessing an index.


orestrnp:
@crayment  pyyyyyinnng

crayment:
Something broke here. This PR used to be small but now contains a bunch of previous commits. Needs fix.

PR #866, opened 2014-07-03
The normal map is now a sprite frame and CCSprite has one too

thayerandrews:
- Change CCEffectRefraction's normal map from CCTexture to CCSpriteFrame.
- Add a normal map sprite frame to CCSprite.
- Promote the normal map to the default uniform set in CCShader.
- Change the GLSL for CCEffectRefraction to index the normal map with cc_FragTexCoord2.
- Add the plumbing needed to populate CCVertex.texCoord2 and cc_FragTexCoord2 from the normal map sprite frame. This comes from the sprite's normal map by default but can be overridden with the effect's normal map if it's present.
- Modify CCEffect's shader uniform management slightly to support all of this. 
- Update CCEffectsTest to set a normal map on the sprite and on the effect to test both paths.


cocos2d/CCSprite.m
```diff
@@ -274,17 +280,18 @@ -(void) setTextureRect:(CGRect)rect rotated:(BOOL)rotated untrimmedSize:(CGSize)
 	_vertexExtents = GLKVector2Make((x2 - x1)*0.5f, (y2 - y1)*0.5f);
 }
 
--(void) setTextureCoords:(CGRect)rect
++ (CCSpriteTexCoordSet)textureCoordsForTexture:(CCTexture *)texture withRect:(CGRect)rect rotated:(BOOL)rotated xFlipped:(BOOL)flipX yFlipped:(BOOL)flipY
```
slembcke, line 283: Ah good. I've been thinking about making this change for a while, but was never needed since the second texture coordinate set hasn't been used for anything official yet.

cocos2d/CCSprite.m
```diff
@@ -498,13 +522,27 @@ -(void) setSpriteFrame:(CCSpriteFrame*)frame
...
+{
+    CCSpriteTexCoordSet texCoords = [CCSprite textureCoordsForTexture:frame.texture withRect:frame.rect rotated:frame.rotated xFlipped:_flipX yFlipped:_flipY];
+    _verts.bl.texCoord2 = texCoords.bl;
+    _verts.br.texCoord2 = texCoords.br;
+    _verts.tr.texCoord2 = texCoords.tr;
+    _verts.tl.texCoord2 = texCoords.tl;
+    
+    // Set the main texture in the uniforms dictionary (if the dictionary exists).
+    self.shaderUniforms[CCShaderUniformNormalMapTexture] = (frame.texture ?: [CCTexture none]);
+    _renderState = nil;
```
slembcke, line 540: You don't have to explicitly reset the render state. CCNode.shaderUniforms handles that for you if adding a custom uniform dictionary invalidates the existing render state.

cocos2d/CCSprite.m
thayerandrews, line 283: Great. I'm glad to know this is inline with your thoughts. The one additional change I considered was to put the new method into some utility or helper class / namespace / file instead of keeping it in CCSprite. I left it in CCSprite for now until there's a strong argument to move it.

cocos2d/CCSprite.m
thayerandrews, line 540: Good to know. I'll add an item to my todo list to remove the render state resets.

PR #869, opened 2014-07-05
Fixes crash when action is getting applied to a deallocated target

videolist:
Making _target property __weak makes sure the pointer is nilled out by ARC when the _target is deallocated


cocojoe:
Thanks, although this is patched in #854

PR #871, opened 2014-07-08
Added properties that can be used to adjust bloom/blur effects.

osinoleg:
Added properties that can be used to adjust bloom/blur effects.

### DIFF
--- a/cocos2d/CCEffectBloom.m
+++ b/cocos2d/CCEffectBloom.m
@@ -67,11 +67,8 @@
 
 -(id)initWithPixelBlurRadius:(NSUInteger)blurRadius intensity:(float)intensity luminanceThreshold:(float)luminanceThreshold
 {
-    _intensity = clampf(intensity, 0.0f, 1.0f);
-    _intensity = 1.0f - _intensity;
-    
-    _luminanceThreshold = clampf(luminanceThreshold, 0.0f, 1.0f);
-    
+    self.intensity = intensity;
+    self.luminanceThreshold = luminanceThreshold;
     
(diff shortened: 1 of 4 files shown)


cocos2d/CCEffectGaussianBlur.h
```diff
@@ -11,8 +11,13 @@
 #if CC_ENABLE_EXPERIMENTAL_EFFECTS
 @interface CCEffectGaussianBlur : CCEffect
 
-@property (nonatomic) GLKVector2 blurDirection;
+// blurRadius number of pixels blur will extend to (6 is the maximum, because we are limited by the number
```
slembcke, line 14: Copy paste error from the comment below I assume?

osinoleg, line 14: yep :P

cocos2d/CCEffectBloom.m
```diff
@@ -189,7 +203,7 @@ -(void)buildFragmentFunctions
     
     // Choose one?
     NSString* addativeBlending =  @"src + dst";
-    NSString* screenBlending = [NSString stringWithFormat:@"(src + dst) - ((src * dst) * %f)", _intensity];
+    NSString* screenBlending = @"(src + dst) - ((src * dst) * u_intensity)";
```
slembcke, line 206: If you added a min(src, dst) blending mode you could have a "gloom" filter too.

osinoleg, line 206: ooo, good idea!! :) I like having a whole new effect, with minimal effort. I'll throw that in with the next pull request.

PR #876, opened 2014-07-09
CCEffectRefraction - Fix various issues uncovered by Viktor during testing

thayerandrews:
- Use premultiplied alpha blending to combine the sprite's main texture with the refracted environment.
- If the refracted texture coordinates would be out of bounds of the environment, do not combine the resulting environment pixel with the sprite's main texture.
- Fix a crash that occurred if an effect was applied to a sprite with a nil main texture.
- If a sprite has no main texture, set the texture rect from the normal map instead.

### DIFF
--- a/cocos2d/CCEffectRefraction.m
+++ b/cocos2d/CCEffectRefraction.m
@@ -74,7 +74,12 @@ static GLKMatrix4 GLKMatrix4FromAffineTransform(CGAffineTransform at);
                                    vec2 refractTexCoords = envSpaceTexCoords.xy + normal.xy * u_refraction;
 
                                    vec4 primaryColor = cc_FragColor * texture2D(cc_MainTexture, cc_FragTexCoord1);
-                                   return primaryColor * texture2D(u_envMap, refractTexCoords);
+                                   if ((refractTexCoords.x >= 0.0) && (refractTexCoords.x <= 1.0) &&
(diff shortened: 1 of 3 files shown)


cocos2d/CCEffectRefraction.m
```diff
@@ -74,7 +74,12 @@ -(void)buildFragmentFunctions
                                    vec2 refractTexCoords = envSpaceTexCoords.xy + normal.xy * u_refraction;
 
                                    vec4 primaryColor = cc_FragColor * texture2D(cc_MainTexture, cc_FragTexCoord1);
-                                   return primaryColor * texture2D(u_envMap, refractTexCoords);
+                                   if ((refractTexCoords.x >= 0.0) && (refractTexCoords.x <= 1.0) &&
```
slembcke: I'm always a little leery of using conditionals in fragment shaders. On pre-A7 devices I've done something like this:

```
vec2 compare = 0.5 - abs(refractTexCoords - 0.5)
primaryColor += step(min(compare.x, compare.y), 0.0)*texture2D(...)
```

thayerandrews: Nice avoidance of the if statement. And I agree I'd rather not introduce the conditional if it can be avoided which apparently it can quite easily. :)

thayerandrews:
Updated to include Scott's optimization above. The only difference is that GLSL's step() function is step(edge, value_to_compare_to_edge) so the args to that function are swapped when compared to Scott's example.

PR #879, opened 2014-07-10
Fix effects in OSX project. Add CCEffectHue.

thayerandrews:
- Replaces CCEffectGlow with CCEffectBloom.
- Updates image paths in the bloom test to work on OSX.
- Adds CCEffectHue and a test for it. The hue adjustments are specified in degrees ranging from -180 to 180.

### DIFF
--- a/cocos2d/CCEffectHue.m
+++ b/cocos2d/CCEffectHue.m
@@ -7,7 +7,87 @@
 //
 
 #import "CCEffectHue.h"
+#import "CCEffect_Private.h"
+#import "CCRenderer.h"
+#import "CCTexture.h"
+
+#if CC_ENABLE_EXPERIMENTAL_EFFECTS
+static float conditionHue(float hue);
+
+@interface CCEffectHue ()
+
+@property (nonatomic) float conditionedHue;
+
+@end
+
 
 @implementation CCEffectHue
 
+-(id)init
+{
+    return [self initWithHue:0.0f];
+}
+
+-(id)initWithHue:(float)hue
+{
+    CCEffectUniform* uniformHue = [CCEffectUniform uniform:@"float" name:@"u_hue" value:[NSNumber numberWithFloat:hue]];
+    
+    if((self = [super initWithFragmentUniforms:@[uniformHue] vertexUniforms:nil varying:nil]))
+    {
+        _hue = hue;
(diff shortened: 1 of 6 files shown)


cocos2d/CCEffectHue.m
```diff
@@ -0,0 +1,160 @@
...
+                                   const highp vec4  kYIQToR      = vec4 (1.0, 0.9563, 0.6210, 0.0);
+                                   const highp vec4  kYIQToG      = vec4 (1.0, -0.2721, -0.6474, 0.0);
+                                   const highp vec4  kYIQToB      = vec4 (1.0, -1.1070, 1.7046, 0.0);
+                                   
+                                   // Convert to YIQ
+                                   highp float YPrime = dot (inputValue, kRGBToYPrime);
+                                   highp float I      = dot (inputValue, kRGBToI);
+                                   highp float Q      = dot (inputValue, kRGBToQ);
+                                   
+                                   // Calculate the hue and chroma
```
slembcke, line 105: You could apply a 2x2 rotation matrix calculated on the CPU here instead and save a lot of expensive trig calls. Actually... since the RGB <--> YIQ conversion is linear, could it be combined into a single color matrix? `(YIQ_to_RGB * rotation * RGB_to_YIQ) * inputValue.rgb`

slembcke, line 105: Looks like yes. I was wondering how that would avoid oversaturation, but you can only avoid it if you use a non-linear color space like HSV (which is too expensive for an effect shader anyway). In the past, I've just rotated around (1,1,1) and called it good enough. Transforming into a proper color space first looks like it helps a lot.

Can even make it a 4x4 transform and have the alpha pass-through. Handy.

thayerandrews, line 105: Neat. I'll work up a pure matrix math version of this and see how the two compare on a device. 

One question: when you say you rotated around (1,1,1) I assume that's in RGB space? Looking at the RGB color cube I can see how that would be roughly equivalent to hue rotation in HSV.

slembcke, line 105: Yes, RGB(1,1,1). It works okay if you don't care about preserving luminance. It's very simple if luminance is not important anyway.

slembcke:
Otherwise looks fine. The color matrix can be added later.

PR #880, opened 2014-07-10
Update CCEffectHue with color matrix implementation. Fix compiler warnings.

thayerandrews:
- As Scott pointed out in his review of my previous pull request, the hue effect could be greatly optimized by using a 4x4 color matrix to do the RGB to YIQ conversion, rotation, and conversion back to RGB. With this optimization, performance of the hue effect test on my iPhone 4s increased from 26fps to 60fps. The original implementation is still in place and disabled with a #if but I will remove it in the future after more testing of the color matrix implementation.
- Remove a bunch of strong reference cycles in the effect pass blocks.
- Fix other miscellaneous compiler warnings from the OSX build.


cocos2d/CCEffectHue.m
```diff
@@ -157,4 +192,32 @@ float conditionHue(float hue)
...
+    // And here:
+    //   http://en.wikipedia.org/wiki/YIQ
+    //
+    // Note that GL matrices are column major so our matrices are transposed relative to the
+    // reference implementations when loaded as they are below. I'm leaving them like this to
+    // improve readability for comparison purposes and I handle the transpose by doing color * mtx
+    // in the shader instead of mtx * color.  The order that the matrices are composed below is
+    // also reversed to account for the left multiply of the color value.
+    
+    GLKMatrix4 rgbToYiq = GLKMatrix4Make(0.299,     0.587,     0.114,    0.0,
```
slembcke: This is a little pedantic, but why not use GLKMatrix4MakeAndTranspose() instead? Then you don't have to reverse the multiplication order nor mix transposed and non-transposed matrices. Not technically wrong, but it does make it non-obvious to follow along without the comment explaining that you purposefully transposed the color transform matrices.

Also, since the untransposed rotation matrix is implicitly inverted as a special case for y/z rotations, I wonder if the hue angle should be negated or not? Especially with the comment in the old code that questioned if the hue was backwards or not.

thayerandrews: Haha. My initial implementation of the color matrix used GLKMatrix4MakeAndTranspose() instead and everything else was reversed. I'll switch it back.

slembcke: :p Ok. I was kind of hesitant to bring it up since it's sort of a dumb nitpick.

thayerandrews: Not at all. I went back and forth trying to figure out which way would be more obvious and in the end I think I basically chose at random. 

You're probably right that creating the matrices with GLKMatrix4MakeAndTranspose() is clearer since it keeps all the matrix operations in a more expected order (at least I feel like v' = Mv is more "standard" than v' = vM) and it puts the word "transpose" right there at matrix creation time.

thayerandrews: Oh also, the sign of the hue rotation is interesting. If you look at this:

http://upload.wikimedia.org/wikipedia/commons/8/82/YIQ_IQ_plane.svg

And this:

http://upload.wikimedia.org/wikipedia/commons/5/52/HSL-HSV_hue_and_chroma.svg

You can see that rotating in the positive direction in YIQ is the opposite of positive hue rotation in HSV. To double check this I wrote a quick test that does the following conversions:
- RGB directly to HSV using the method described [here](http://en.wikipedia.org/wiki/HSL_and_HSV#General_approach).
- RGB to YIQ using the matrix from the code.
- Compute H from YIQ as the GPUImage source does.

```
rgb:( 1.000  0.000  0.000)  yiq:( 0.299  0.596  0.211)  hsv:(   0.000  1.000  1.000)  H from YIQ: 19.543 
rgb:( 1.000  1.000  0.000)  yiq:( 0.886  0.321 -0.311)  hsv:(  60.000  1.000  1.000)  H from YIQ:-44.082
rgb:( 0.000  1.000  0.000)  yiq:( 0.587 -0.274 -0.523)  hsv:( 120.000  1.000  1.000)  H from YIQ:-117.707
rgb:( 0.000  1.000  1.000)  yiq:( 0.701 -0.596 -0.211)  hsv:( 180.000  1.000  1.000)  H from YIQ:-160.475
rgb:( 0.000  0.000  1.000)  yiq:( 0.114 -0.321  0.311)  hsv:( 240.000  1.000  1.000)  H from YIQ: 135.898
rgb:( 1.000  0.000  1.000)  yiq:( 0.413  0.274  0.523)  hsv:( -60.000  1.000  1.000)  H from YIQ: 62.302
```

From this you can see that the H values from YIQ match the image of YIQ space but they don't correspond to H angles for HSV and, most importantly for this discussion, they are backwards relative to one another.

thayerandrews:
Updated CCEffectHue to address Scott's comments. This also includes a change to CCEffectRefraction to address Viktor's request for masking refraction based on the normal map's alpha channel.

slembcke:
Looks good to me. No more matrix nitpicks. ;)

PR #882, opened 2014-07-11
Only forward touch events if CCResponder has user interaction enabled

Ben-G:
With the current implementation touch events caused by an active touch will be forwarded to a `CCResponder` even if the responder sets `userInteractionEnabled` to `NO`. 

Can be reproduced with the following sequence:
- Start touch
- Set `userInteractionEnabled`to `NO`
- Move touch

With the current implementation `touchMoved` would be called in the above example. 
With the suggested change a touch event will only be forwarded if a node has `userInteractionEnabled` set to `YES`.

### DIFF
--- a/cocos2d/CCResponderManager.m
+++ b/cocos2d/CCResponderManager.m
@@ -433,7 +433,13 @@
 {
     for (CCRunningResponder *touchEntry in _runningResponderList)
     {
-        if (touchEntry.touch == touch) return(touchEntry);
+        if (touchEntry.touch == touch) {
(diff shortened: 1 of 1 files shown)


Birkemose:
I think, that if the user disables touch handling while a touch is ongoing, there are two options
1) complete the touch, as the user has already accepted it (current functionality)
2) forcefully cancel touch
I don’t think just stopping to respond to an ongoing touch, is the right way.

Ben-G:
Yes I was thinking about cancelling the touch as well. I would be fine with solution 2). Any other opinions on option 1) vs option 2)?

PR #886, opened 2014-07-14
Fixed bug #802. Changed opacity value from unsigned char to float.

chadwyk:
Fixed bug #802. Changed opacity value from unsigned char to float.

http://forum.cocos2d-iphone.org/t/3-1-rc1-tilemap-layer-opacity-problem/13782


slembcke:
Thanks!

PR #897, opened 2014-07-21
Refraction, reflection, and glass

thayerandrews:
- Addresses the remainder of Viktor's refraction feedback by establishing a range for the refraction value. -1 is maximum shrinking of the refraction image, 0.0 is no change, 1.0 is maximum magnification.
- Add CCEffectReflection.
- Add CCEffectGlass which combines both refraction and reflection.
- Address a few low hanging performance issues in CCEffects that were uncovered by Instruments.


cocos2d/CCEffectRenderer.m
```diff
@@ -19,6 +19,16 @@
 
 #if CC_ENABLE_EXPERIMENTAL_EFFECTS
 
+#define CCEFFECT_DEBUG_PASSES 0
+#if CCEFFECT_DEBUG_PASSES
+#define CCEFFECT_DEBUG_PUSH_GROUP(R) [(R) pushGroup]
+#define CCEFFECT_DEBUG_POP_GROUP(R, L) [(R) popGroupWithDebugLabel:(L) globalSortOrder:0]
+#else
+#define CCEFFECT_DEBUG_PUSH_GROUP(R)
```
slembcke: Skipping the push/pop of the group for debug purposes isn't quite right. The blur effects break for sure. Did you just want to skip the debug label generation?

slembcke:
Looks neat. Except for the push/pop bug it looks fine.

Don't worry about it for this request (yours are okay), but do be a little careful with testing asset file sizes. A few months ago we rewrote the repository history to slim it down so it wasn't such a pain to use as a submodule. One of the smaller, but significant things removed was ~50MB of old unused testing assets.

cocos2d/CCEffectRenderer.m
thayerandrews: Okay. The way I remembered it was that I introduced the push/pop to aid in debugging the corresponding GL calls but maybe the push/pop was already there. First I fixed debug label creation (doing it ahead of time and storing it in CCEffectRenderPass instead of doing it repeatedly on the fly) but then I still saw that push/pop was fairly expensive so I switched to the macros.

Thanks for pointing out the issue. I'll dig further into it.

thayerandrews:
The pushGroup / popGroup change is rolled back. Thank you for catching that.

slembcke:
Push/pop can create a lot of array garbage if there are a lot of calls to them. It wasn't a priority in the past, but this is the second that pooling the queues has come up, so I'll bump up the priority. ;)

jtwigg:
@slembcke If this ones ready to go, can we get it pull'd? I'm working on the SB UI and waiting for these changes.

Cheers.

PR #920, opened 2014-08-06
CCEffectNode - Fix a bunch of sizing issues and add tests for them

thayerandrews:
- Refactor CCRenderTexture's create method into create and createTextureAndFboWithPixelSize. In CCEffectNode, override the create method so pixel size is computed correctly. This fixes effect node's parent relative sizing while leaving render texture the same (because render texture needs even more changes that I don't want to tackle right now).
- Override setContentSizeType on effect node so it also marks the content size as dirty when it changes.
- If content size has changed since the last visit, destroy any allocated resources and recreate them (as CCRenderTexture already does).
- Pass the content size in points to CCEffectRenderer from sprite and effect node.
- Add a bunch of effect node sizing tests.

**Update**:
- Fix handling of sprite color and opacity when combined with effects.
- Fix handling of clipping nodes as children of effect nodes.
- Fix effect node resizing when an effect node's size is parent relative and the size of a parent changes.
- Add tests for all of this.

### DIFF
--- a/cocos2d/CCRenderTexture.m
+++ b/cocos2d/CCRenderTexture.m
@@ -169,12 +169,14 @@
 
 -(void)create
 {
+    CGSize pixelSize = CGSizeMake(_contentSize.width * _contentScale, _contentSize.height * _contentScale);
+    [self createTextureAndFboWithPixelSize:pixelSize];
+}
+
+-(void)createTextureAndFboWithPixelSize:(CGSize)pixelSize
+{
 	glPushGroupMarkerEXT(0, "CCRenderTexture: Create");
 	
-	int pixelW = _contentSize.width*_contentScale;
-	int pixelH = _contentSize.height*_contentScale;
-
-
 	glGetIntegerv(GL_FRAMEBUFFER_BINDING, &_oldFBO);
 
 	// textures must be power of two
@@ -182,16 +184,16 @@
 	NSUInteger powH;
 
 	if( [[CCConfiguration sharedConfiguration] supportsNPOT] ) {
-		powW = pixelW;
-		powH = pixelH;
+		powW = pixelSize.width;
+		powH = pixelSize.height;
 	} else {
-		powW = CCNextPOT(pixelW);
-		powH = CCNextPOT(pixelH);
+		powW = CCNextPOT(pixelSize.width);
+		powH = CCNextPOT(pixelSize.height);
 	}
 
 	void *data = calloc(powW*powH, 4);
 
-	CCTexture *texture = [[CCTexture alloc] initWithData:data pixelFormat:_pixelFormat pixelsWide:powW pixelsHigh:powH contentSizeInPixels:CGSizeMake(pixelW, pixelH) contentScale:_contentScale];
+	CCTexture *texture = [[CCTexture alloc] initWithData:data pixelFormat:_pixelFormat pixelsWide:powW pixelsHigh:powH contentSizeInPixels:pixelSize contentScale:_contentScale];
     self.texture = texture;
     
 	free(data);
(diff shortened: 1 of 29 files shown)


cocos2d/CCRenderTexture.m
```diff
@@ -169,12 +169,14 @@ -(id)init
 
 -(void)create
 {
+    CGSize pixelSize = CGSizeMake(_contentSize.width * _contentScale, _contentSize.height * _contentScale);
```
slembcke, line 172: Should be self.contentSizeInPoints.width \* _contentScale ?

cocos2d/CCRenderTexture.m
```diff
@@ -234,6 +236,10 @@ -(void)create
 	CC_CHECK_GL_ERROR_DEBUG();
 	glPopGroupMarkerEXT();
 	
+    // XXX Thayer says: I think this is incorrect for any situations where the content
```
slembcke, line 239: Thought about this a couple months ago and was sort of avoiding it until it was a problem. If the content size is parent relative then the content size in points can change at any time (even between frames) without any warning or event that can be caught. While somebody might set up a render texture this way, I'm not even sure if there is anything remotely correct we can do.

When put that way, I'm wondering if the only correct thing we can do is to make it an error to use anything but points for the content size type.

cocos2d/CCRenderTexture.m
thayerandrews, line 172: Yup. This should be fixed as well but I was leaving it out of this set of commits since I only wanted to change CCEffectNode here.

cocos2d/CCRenderTexture.m
thayerandrews, line 239: I see what you mean. If the content size type is parent relative, is it possible to determine the correct size during visit and compare this to the size that was last used to alloc the texture / FBO? If the two aren't equal then we destroy and recreate the texture and FBO prior to rendering the render texture / effect node's children? 

I'm guessing there's a situation (or two or three) that I'm overlooking where this won't work. And it would also mean comparing the sizes every frame to see if the texture reallocation needs to happen.

slembcke, line 239: The problem is that the "auto-draw" functionality is pretty rarely used I think. Personally, I never noticed it until I was rewriting the class for 3.1. Before that I'd never heard it come up in the forums for instance. CCRenderTexture is primarily used by calling the begin/end methods explicitly from outside of a draw method, usually in update, and often only when content changes. Deferring the resize would mean that it have to invalidate the content the user just drew into it.

SpriteBuilder doesn't let you explicitly set the content size/type of a CCSprite. I'm not really sure what would happen if you did it in code. In both cases, I think it makes very little or no sense to allow parent relative content sizes at all.

thayerandrews, line 239: Okay. That makes sense for CCRenderTexture. CCEffectNode actually only has auto-draw so I think the "check the current dimensions against the texture dimensions" fix would work there no?

For CCRenderTexture, could we add another mode so the options are:
- Manual draw
- Auto draw
- Manual draw but auto-draw if it is determined that the dimensions are invalid.

Or is it better to just leave CCRenderTexture along until version 4 when we have more leeway to change its behavior?

slembcke, line 239: I'm leaning towards leaving it alone until v4. There are already a number of other problems that I don't really know what to do about without breaking the API or making it even more complicated. SpriteBuilder doesn't expose render textures (yet), and I don't know of any issues that have come up related to this yet.

thayerandrews, line 239: Sounds good. I won't make any changes to CCRenderTexture and I'll talk to Viktor about whether or not he wants to keep these changes to CCEffectNode.

PR #945, opened 2014-08-21
Instant fix

JoshBoluga:
CCActionEaseInstant's timing is wrong and calls update:1 when t is not less than 0. This means the action goes to the end immediately. The instant change should happen at the end. This is particularly noticeable if you're using SpriteBuilder, as your keyframes are applied at the beginning of the transition – at the time of the previous keyframe.

### DIFF
--- a/cocos2d/CCActionEase.m
+++ b/cocos2d/CCActionEase.m
@@ -460,7 +460,7 @@
 @implementation CCActionEaseInstant
 -(void) update: (CCTime) t
 {
-    if (t < 0)
+    if (t < 1)
     {
         [self.inner update:0];
     }


cocojoe:
I would like to check the ease against a couple of SB projects first, just in case.

JoshBoluga:
That sounds advisable as I've only tested it against my own projects.

This is my first pull request, and I'm a bit confused as to why GitHub's saying there's two commits. I only made the one-line (actually only one character!) commit about the CCActionEaseInstant change.

cocojoe:
Josh presume you were testing with latest Cocos2D develop? As there were a few changes in last few weeks for animation and in particular instant easing.

JoshBoluga:
I may have missed it. Sorry!

cocojoe:
Oh I ment in CCAnimationManager, your change is valid, I tested and was fine.

JoshBoluga:
Great! Reopened.

JoshBoluga:
Is there any reason why this issue has been closed? The issue appears to remain in code.

andykorth:
So I created a handful of unit tests for the purposes of testing this and other ease actions. It's important to note that CCActionEaseInstant is a CCActionEase, not a CCActionInstant... so it's actual intended behavior is a little fuzzy. It does have a duration, which it picks up from the attached action. 

Most actions, including EaseIn and EaseOut, have special logic to ignore the first time step that occurs and not increase the elapsed time. CCActionEaseInstant is special in this case, because even if the [update method](https://github.com/JoshBoluga/cocos2d-iphone/commit/20e4633b69d0de984293646b876ec20993d165ef#diff-e4d7ae24c7b3095aa63f73bbb78943e6L463) is run with a elapsed time of zero, the else statement is evaluated and the inner action is called with an elapsed time of 1. 

I doubt the current behavior was intended (perhaps the fix should be changing the `t < 0` to `t <= 0`). But the current code definitely completes the action the first time it's stepped, without ever evaluating it with a time of zero. If the current behavior was intended, the code would probably look like this:

```
@implementation CCActionEaseInstant
-(void) update: (CCTime) t
{
    [self.inner update:1];
}
@end
```

However, I think this [commit](https://github.com/JoshBoluga/cocos2d-iphone/commit/20e4633b69d0de984293646b876ec20993d165e) would have some undesired behavior. The "instant" action would not occur at all until the action was stepped enough times to advance through its entire duration. Basically, the duration would be treated as a delay before the entire action occurs, and the inner action would be evaluated potentially many times with a duration of zero.

I propose this change instead:

```
@implementation CCActionEaseInstant
-(void) update: (CCTime) t
{
    if (t <= 0)
    {
        [self.inner update:0];
    }
    else
    {
        [self.inner update:1];
    }
}
@end
```

This ensures the inner action will be evaluated once at t=0, then on the next step it will be evaluated at t=1. Some actions seem to rely on being evaluated at t=0, but that's something to reconsider later.

This solution should prevent the Spritebuilder issues of actions evaluating on their t=0 keyframe.

andykorth:
My new unit tests are here:
https://github.com/andykorth/cocos2d-iphone/blob/1ee7d4a65fa9f12a569b7ffea6236a45623b13e2/UnitTests/CCActionTest.m#L154

There's a pull request to merge them into v4.

Birkemose:
I have always wanted the actions to expose a normalised transfer function, for the user to override, or provide a selector for, so that you easily can customise the behaviour.

andykorth:
Yeah, CCActionCallFunc operates on selectors and CCActionCallBlock is even better- it operates on blocks. I've written things similar to CCAction in other engines, and I found that calling a block is the only action I've ever needed. Updating a position in a block is easier than remembering CCActionMoveTo, for example. But.... we've already got a few thousand lines of single-use actions :P

andykorth:
Oh yeah, and I think a similar method could be used for custom ease functions too. That's a good idea!

PR #948, opened 2014-08-23
Use autorelease CCRenderState method in CCLabelTTF

federicobond:
This fixes a build error introduced in f52a39197f

```
cocos2d-iphone/cocos2d/CCLabelTTF.m:340:41: No visible @interface for 'CCRenderState' declares the selector 'initWithBlendMode:shader:shaderUniforms:copyUniforms:'
```


slembcke:
Oof. Not sure how I missed that. Merging into too many branches... Thanks.

PR #953, opened 2014-08-28
Handle odd line seperators in CCLabelBMFont.

sarah-j-smith:
The XCode plist editor and some other editors can create string data
with Unicode line seperators in it, eg: \u2028.

http://en.wikipedia.org/wiki/Newline#Unicode

You can easily enter one using <ctrl>+<enter> in your plist or code
editor.  The plist editor will not let you type <enter> as it will just
commit the current cell being edited.  To get an actual newline you can
type <option>+<enter>.

Code in CCLabelBMFont correctly tests using NSCharacterSet
newlineCharacterSet in most places, but in two places incorrectly
tests for equality against '\n'.

So strings with one of these "odd" newlines will be rendered with all
the text on one line.  Fix this, and make the code consistent.


slembcke:
Sorry it took a while to get this reviewed!

PR #955, opened 2014-08-28
Tween compatible instant easing fix

cocojoe:
Tween uses a NULL keyframe, so instant ease code was being called as instant ease enum is 0.


cocojoe:
Will redo this for 3.2

PR #958, opened 2014-08-30
Added alternative reschedule:interval method to CCNode class.

ghost:
Added a reschedule:interval method to the CCNode class. Passing in a selector that is already scheduled to this method will update the interval time but won't unschedule and reschedule the method like calling schedule:interval does.


slembcke:
Looks fine to me.

PR #963, opened 2014-09-03
CCEffects - Add color channel offset effect, fix effect clipping artifacts

thayerandrews:
- Add CCEffectColorChannelOffset which allows the red, green, and blue color channels to be offset from one another by independent 2D position vectors.
- Fix clipping of effects like blur that increase the size of the affected sprite. The user can now specify padding for their effects that increases the size of the submitted geometry without changing the visible size of the sprite image. Enlarging the geometry adds scratch space around the sprite image so blurs, blooms, and other effects do not get clipped by the geometry edges.
- Update the OSX project to include all the new effects.


slembcke:
Looks ok, but I can't run the tests. Did you forget to commit `Images/WhiteDiamond.png`?

thayerandrews:
Thanks for catching that. I don't actually want to add WhiteDiamong.png since I'm trying to avoid adding new test images if I don't need to. I'll update the test to use an existing file from the Images folder.

thayerandrews:
Okay. The tests should be fixed.

thayerandrews:
**Update:**
- Fix padding support for effects like Bloom that use different texture coordinate sets to index the previous pass texture and the main texture.
- Fix padding support for effects that do not require padding themselves but have to account for padding when they are stacked with effects that do need it.
- Add more padding test cases.

cocos2d/CCEffectBloom.m
```diff
@@ -205,19 +217,27 @@ -(void)buildFragmentFunctions
             GLfloat optimizedWeight = firstWeight + secondWeight;
             GLfloat optimizedOffset = (firstWeight * (currentOverlowTextureRead * 2 + 1) + secondWeight * (currentOverlowTextureRead * 2 + 2)) / optimizedWeight;
 
-            [shaderString appendFormat:@"srcPixel = texture2D(cc_PreviousPassTexture, v_blurCoordinates[0] + singleStepOffset * %f);\n", optimizedOffset];
+            [shaderString appendFormat:@"blurCoords = v_blurCoordinates[0] + singleStepOffset * %f);\n", optimizedOffset];
```
slembcke, line 220: This line outputs shader code with an unmatched parenthesis:

```
blurCoords = v_blurCoordinates[0] + singleStepOffset * 7.384912);
```

cocos2d/CCEffectBloom.m
```diff
@@ -205,19 +217,27 @@ -(void)buildFragmentFunctions
             GLfloat optimizedWeight = firstWeight + secondWeight;
```
slembcke, line 217: Does it make sense to have separate independent/dependent texture read paths now that they are always dependent?

slembcke:
Looks like there is a minor issue with the bloom effect left.

slembcke:
Eh, ignore me. Had I looked at it for 10 more seconds before posting I would have come to the conclusion that the extra parenthesis should be removed. Will close this shortly.

cocos2d/CCEffectBloom.m
thayerandrews, line 217: This is something I've been wondering about myself. I think we should test
the performance of the two different paths and see if it's worth
maintaining different implementations.

PR #972, opened 2014-09-06
fixing CCRenderTexture -clearStencil

ksjogo:
I did not know how to exactly test this, since nobody is calling that in core, but it reads wrong the way it is.


slembcke:
/me facepalm

Copy paste error in a method I suspect nobody has ever used. -_-

PR #983, opened 2014-09-17
Feature/packages

NickyWeber:
Packages support for cocos2d.


slembcke:
Err, yes the CCFileUtils issue was my fault. Didn't check compiling for all platforms before pushing.

PR #984, opened 2014-09-17
CcNodeColor cascade opacity

RobBoluga:
Fix to make opacity from the parent CCNode cascade to CCNodeColor and CCNodeGradient

### DIFF
--- a/cocos2d/CCNodeColor.m
+++ b/cocos2d/CCNodeColor.m
@@ -192,7 +192,6 @@
 	
 	float gmin = MIN(MIN(g0, g1), MIN(g2, g3));
 	float gmax = MAX(MAX(g0, g1), MAX(g2, g3));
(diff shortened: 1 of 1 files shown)


cocos2d/CCNodeColor.m
```diff
@@ -237,6 +243,12 @@ -(void) setEndOpacity: (CGFloat) o
 	[self updateColor];
 }
 
+- (void)updateDisplayedOpacity:(CGFloat)parentOpacity
```
slembcke: Duplicate? It will inherit this from CCNodeColor no?

RobBoluga: Sorry, yes, you are correct.
I’ll update with it removed.

Rob

slembcke:
Thanks!

PR #986, opened 2014-09-17
Stop the CCDirector from running when the app is backgrounded

sridenour:
Stop the CCDirector from running when the app is backgrounded, regardless of what view controller is visible (because it will still crash if it tries to render in the background when it isn't the "visible" view controller).


slembcke:
Hadn't merged the 3.x branches forward yet, but it's been done now.

Issue #973, opened 2014-09-09
Portrait orientation does not work on iOS 8


christinakelly:
Portrait apps are misaligned on iOS 8.  Tested on an iPad Air.

osinoleg:
looking into this now.
- (BOOL)resizeFromLayer:(CAEAGLLayer *)layer 
  is called with a layer.bounds that don't match the window bounds. - only happens on iOS8. I also noticed that if I manually set the desired orientation in the info.plist, then everything seems to work ok.

osinoleg:
So this looks like and SDK bug. I can reproduce this bug with provided sample projects from apple (AdaptivePhotos app located here: https://developer.apple.com/wwdc/resources/sample-code/).
Here are a bunch of people having the same issue: https://devforums.apple.com/message/1044988#1044988

Note: it seems like this issue only happens on iPhone devices. 
Solution: set the desired orientation in the info.plist until apple has a fix.

Another note: If the user builds xcode5 and deploys to a ios8 devices, this bug does not show up,.

osinoleg:
Came up with a hack around this issue, by forcing the status bar orientation. 

Pull request: https://github.com/cocos2d/cocos2d-swift/pull/987

Issue #964, opened 2014-09-03
Tilemap parsing issue in iOS 8.


slembcke:
Tilemap parsing doesn't work in iOS 8. It crashes on the assertion `NSXMLParser does not support reentrant parsing`.

The .tmx file references and attempts to load a .tsx file. The .tsx file is using a unique parser object. That seems like something that should be Ok, but digging into some [old release notes](https://developer.apple.com/library/ios/releasenotes/Foundation/RN-Foundation-iOS/Foundation_iOS5.html) I found this:

> NSXMLParser is now threadsafe. However, it is not reentrant on a given thread; don't call -parse on an NSXMLParser from within a delegate callback of another NSXMLParser.

It's mentioned in the [class docs](https://developer.apple.com/library/prerelease/ios/documentation/Cocoa/Reference/Foundation/Classes/NSXMLParser_Class/index.html) too, but it's not as clear:

> Unless used in a callback, the NSXMLParser is a thread-safe class as long as any given instance is only used in one thread.

So this has been incorrect in Cocos2D for quite some time. It's only in iOS 8 that Apple added an assertion for it. There are a number of people using a workaround where they create a new thread to run the child parser in. We should probably look into that.

slembcke:
I'm also worried that this is going to be a retractive bug. The tilemap parsing code has changed very little for many years. This might affect all Cocos2D apps running on iOS that use tilemaps regardless of what version of Cocos2D they use.

osinoleg:
Fixed, pull request: https://github.com/cocos2d/cocos2d-swift/pull/987

YvesLeBorg:
apparently the same issue on OSX ?  see below

http://stackoverflow.com/questions/26621898/cocos2d-3-2-1-parallax-test-crashes-on-mac-os-x-10-10-xcode-6-1

PR #987, opened 2014-09-18
Bug fixes (#964 #973) - TileMap and Orientation fixes. 

osinoleg:
Make CCXMLParser thread safe (avoid re entry on same thread). github issue: #964

Fix orientation bug for ios8/xcode6 builds (by forcing the orientation of the statusbar, yes this is a weird hack to work around Apple's bug). github issue: #973


cocos2d/CCTMXXMLParser.m
```diff
@@ -252,7 +252,13 @@ -(void)parser:(NSXMLParser *)parser didStartElement:(NSString *)elementName name
 
 			_currentFirstGID = [[attributeDict objectForKey:@"firstgid"] intValue];
 		
-			[self parseXMLFile:externalTilesetFilename];
+            // since the xml parser is not reentrant, we need to invoke each child xml parser in its own queue
+            dispatch_queue_t reentrantAvoidanceQueue = dispatch_queue_create("xmlParserSafeQueue", DISPATCH_QUEUE_SERIAL);
+            dispatch_async(reentrantAvoidanceQueue, ^{
```
slembcke, line 257: Any reason to do both an async and a sync call instead of just a sync call?

PR #990, opened 2014-09-23
Fixes for building and debugging android test project

breckinloggins:
The main issue is that the RELEASE configuration incorrectly specified building Java targets with dSYM support, which the Java compiler obviously does not support.

In addition, this adds the "INTERNET" permission to the AndroidManifest.xml file, which is needed to be able to open a connection on the device side for remote debugging for all devices (the previous method did not require this, but also did not work with devices that had problems with the run-as command).

Note that there are currently issues on the SDK side with running apps built in RELEASE mode. I'll work on a fix for that separately. For now, please continue to build the debug test app in DEBUG mode.


breckinloggins:
This was developed and tested on Xcode 6.0.1 (6A317) but should work on Xcode 5.1.1 as well.

PR #1021, opened 2014-10-09
Combined projects

slembcke:
Contains the following merges:
https://github.com/cocos2d/cocos2d-swift/pull/1019
https://github.com/cocos2d/cocos2d-swift/pull/1020

This cannot be merged until @collinjackson's ObjectAL pull request is merged here:
https://github.com/spritebuilder/ObjectAL-for-Cocos2D/pull/2

Also combines the Mac/iOS ui-tests projects into a single project so we have many fewer duplicate projects floating around instead of one more. Still working on adding an Android ui-test target. This should make testing on all of the platforms a lot less painful in the future hopefully!


UnitTests/CCUnitTestAssertions.h
```diff
@@ -1,15 +1,19 @@
...
-        _XCTRegisterFailure(_XCTFailureDescription(_XCTAssertion_EqualObjects, 0, @#a1, @#a2, a1value, a2value),format); \
-    } \
-    @catch (id exception) { \
-        _XCTRegisterFailure(_XCTFailureDescription(_XCTAssertion_EqualObjects, 1, @#a1, @#a2, [exception reason]),format); \
-    } \
-} while(0)
\ No newline at end of file
+#define CCAssertEqualStrings(a1, a2, format...) XCTAssertEqualObjects(a1, a2, format)
+
+// TODO had to replace this macro due to compile errors in calling _XCTRegisterFailure.
```
slembcke, line 3: @NickyWeber Heads up that I had to temporarily replace your string assert macro to get the unit tests to compile. Spent a while trying to debug the macro but gave up.

collinjackson:
I merged https://github.com/spritebuilder/ObjectAL-for-Cocos2D/pull/2

slembcke:
Fixing (hopefully) one last resource issue. Didn't do a clean enough clean build last time.

slembcke:
Just a heads up, people should run `git clean -fd .` after this is merged since the old project and workspace files contain ignored files and won't be automatically removed by Git.

UnitTests/CCUnitTestAssertions.h
NickyWeber, line 3: @slembcke What's the status on the macro? I can fix it. Macros can be such a pain.

slembcke, line 3: Ah. So figured out what the problem was. _XCTRegisterFailure() takes an extra parameter now for the current test case. Looks like those implementation macros were supposed to be private and changed in the new SDK.

Is the following an acceptable replacement?

```
#define CCAssertEqualStrings(a1, a2, format...) ({ \
    id a1value = (a1); \
    id a2value = (a2); \
    XCTAssertTrue([a1value isKindOfClass:[NSString class]], format); \
    XCTAssertTrue([a2value isKindOfClass:[NSString class]], format); \
    XCTAssertEqualObjects(a1value, a2value, format); \
});
```

Not really sure I understood the reasons for why it was implemented the way it was though. (custom assertion vs. XCTAssertEqualObjects(), -compare: vs. -isEqual:)

PR #1056, opened 2014-11-05
Adding rough support for iPhone 6 and 6+.

slembcke:
Sort of a terrible hack, but it's not any less future proof as the existing code. Should work at least until Apple introduces a new screen size or v4.


KristenY:
@slembcke 
Related to the changes introduced by this iPhone 6 and 6+ pull request - could you please take a look at the changes here?  https://github.com/cocos2d/cocos2d-swift/pull/1061

PR #1059, opened 2014-11-07
CCBReader property support.

slembcke:
Make CCBReader try to set a property before directly setting an ivar. Useful, and also bypasses retain counting issues with Swift classes and object_setIvar().


cocojoe:
Great, will test this out right now.

PR #1061, opened 2014-11-08
Compile error resolution for CCGLView.m

KristenY:
Changes from commit https://github.com/cocos2d/cocos2d-swift/commit/dc0a2a892c0f745fcf1d2a657e1d8f672107d580 were causing builds for iOS to fail; this change allows them to build & run.


slembcke:
What SDK version is this failing to compile for? It's defined for 8.0 and up. Using native scale is required to support the iPhone 6+ properly.

vlidholt:
Seems like the reason that it isn't compiling is that an old version of Xcode was used.

PR #1065, opened 2014-11-12
CCShader - Add non-prefixing, raw GLSL compilation path

thayerandrews:
- Rework the flow from initWithVertexShaderSource... -> initWithGLVertexShaderSource -> CompileShader so prefixing is not done at the end of this call chain but at the beginning. CompileShaderSources (CompileShader's new name) now blindly compiles whatever source strings are passed to it without modification.
- Add initWithRawVertexShaderSource... that does not insert any prefix headers into the supplied source.
- Modify compilation of our built-in shaders to use the raw compilation path.
- This also fixes the OS X and Android builds by adding CCLightNode to them.


slembcke:
Sorry for the delay. Looks good to me.

thayerandrews:
No problem. I know you're busy with planning work this week. Thanks for taking a look.

PR #1067, opened 2014-11-13
Fix ivar memory management 

mauiaaron:
Fixes reusing memory after deallocation in Android CCActivity


slembcke:
Eeee. Dangling pointers. Surprised that didn't crash all the time. Thanks.

PR #1078, opened 2014-11-18
v3.3  api reference cleanup pass

LearnCocos2D:
updated ignore list and fixed the more pressing or obvious appledoc warnings


LearnCocos2D:
that was fast, got another one coming ;)

cocojoe:
I did it first time around, so am familiar :)

PR #1088, opened 2014-11-23
Only enable Metal support on 64-bit devices.

sridenour:
Prior to this commit, Cocos2D will still be built and run with Metal support if CC_ENABLE_METAL_RENDERING is 1 and we are building with the iOS 8 SDK, even if we are running on a device that doesn't support Metal but supports iOS 8 such as the iPhone 5. When this happens, Cocos2D crashes.

What really should happen is that if creating the Metal view fails then Cocos2D should fall back to OpenGL, but this will have to do for now.

### DIFF
--- a/cocos2d/ccMacros.h
+++ b/cocos2d/ccMacros.h
@@ -56,7 +56,8 @@
 #endif
 
 // Metal is only supported on iOS devices (currently does not include the simulator) and on iOS 8 and greater.
-#if __CC_PLATFORM_IOS && defined(__IPHONE_8_0) && __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_8_0
+// Metal is only supported on 64-bit devices.
+#if __CC_PLATFORM_IOS && defined(__IPHONE_8_0) && __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_8_0 && defined(__LP64__)
 #define __CC_METAL_SUPPORTED_AND_ENABLED (CC_ENABLE_METAL_RENDERING && !TARGET_IPHONE_SIMULATOR)
 #else
 #define __CC_METAL_SUPPORTED_AND_ENABLED 0


sridenour:
Yes. On my iPhone 5 running iOS 8.1, with Metal enabled in ccConfig.h, it tries to launch Metal, fails, and crashes (Cocos2D Metal support works fine on my iPhone 6). This is with Xcode 6.1. Metal.framework is linked as Optional.

slembcke:
Looks like I need to update my iPad 2 to iOS 8. (I was going to keep it as a 7.x testing device)

I could never find any docs on the "correct" way to detect Metal support at runtime. Checking for the existence of classes or methods is the usual way of detecting runtime support on Apple platforms. I was given the impression that the Metal framework would simply not be available on devices that didn't support it. Apparently that is not the case. -_-

It's unlikely that Apple is going to make a new 32 bit device, or a 64 bit one that doesn't support Metal, but I want to dig into this a little bit more.

sridenour:
I couldn't find docs about detecting Metal support either.

The crash comes as a result of an NSAssert() failure in CCShaders.m on line 712, in the +initialize method. Seems it can't load the Metal shader library, so the assertion fails, throwing an exception.

Maybe that's how the Metal test could be done: test for the existence of Metal (to allow Cocos2D to still run on iOS 7), and if Metal exists see if the shader library can be loaded. If it can't be loaded, revert to OpenGL instead of throwing an exception.

But, like you said, it's unlikely Apple would ship a new 32-bit iDevice going forward, and it's _extremely_ unlikely that Apple will drop OpenGL ES any time soon.

Interestingly, Metal works in the 32-bit environment _on 64-bit devices_. I made a new project in SpriteBuilder, enabled Metal support, then disabled building ARM64 slices (via setting Architectures to ARCHS_STANDARD_32_BIT in build settings) and set "Build active architecture only" to NO, so that it would only build a 32-bit binary. When run on my iPhone 6 this 32-bit binary successfully initialized and used Metal. Cocos2D debug console output confirmed that it was running in the 32-bit environment.

slembcke:
Oops. Forgot to close this a while back.

I looked into it more. Still couldn't find anything in the docs about the recommended way to do it. The framework is installed, and links even on devices that don't support Metal (which I originally did not expect). Functions such as MTLCreateSystemDefaultDevice return `nil` though. So I guess (probably?) the correct thing to do is continue to weak link the framework and check function pointers at runtime, but also check that it can create a device. This seems fairly future proof unless Apple decides to do something really weird.

PR #1091, opened 2014-11-26
viewDidResizeTo should use points instead of pixels

andykorth:
When the view is resized (or device orientation changed), viewDidResizeTo messages are sent to all the CCNodes in the scene. The value sent should be in points, not in pixels.

Sending this to develop-v4 instead of earlier versions, since Scott mentioned this was part of cocos3d and might break some cocos3d behavior.


andykorth:
Interestingly, on Android it was already set to points:

https://github.com/cocos2d/cocos2d-swift/blob/3cee9193f2c69ea8ab7f39c142df005f85cac6e8/cocos2d/Platforms/Android/CCDirectorAndroid.m#L111

Issue #1096, opened 2014-12-02
Mac Spritesheets not using correct folders in CCBReader


cocojoe:
Mac spritesheets not working correctly due to CCBReader folders not being set correctly.  

Related: SpriteBuilder/SpriteBuilder#1192

Issue #1053, opened 2014-11-03
NSEvent.h/.m not added to cocos2d-mac target


cocojoe:
They are very handy for Mac project.

NickyWeber:
Looks good too, but can't merge due to lack of privileges.

Issue #1095, opened 2014-12-01
Mac target NSEvent category missing in Xcode project


ghost:
NSEvent+CC..m is missing in Platforms/OSX.

The category implementation is there
but not part of the Xcode project.

cocojoe:
Oh still?

#1053

PR #1097, opened 2014-12-02
Mac spritesheet fix, Missing mac files

cocojoe:
I used phone/phonehd as they look right and are closest to actual resolutions used.
Quickly fixed Mac NSEvent missing files #1053


cocojoe:
Fixes #1096 & #1095

PR #1100, opened 2014-12-03
Develop v4

slembcke:
This is a pretty huge pull request (sorry about that), though you can probably skim through it 10 minutes or so if you go commit by commit. I would not recommend viewing it all at once since there have been many changes of various types to the same files. I'll try to avoid such huge pull requests in the future. All of these commits were made within days of each other and I've been avoiding making pull requests for them for a week because of the size.

Most of the changes are from cleaning up the headers imports. There were a huge number of circular imports when forward declarations should have been used instead. This was making it nearly impossible to move definitions between headers without causing compile errors in half of the .m files in the project.

Most of the remaining changes are either removing old deprecated functionality, adding new deprecated methods (the old 2.x color/vertex structs, affine transforms, etc) or cleanup of private methods.

If nobody says anything, I'll go ahead and merge this in a few days. There are already a few dozen conflicts with the develop branch after merging Steffen's doc changes and I'd like to avoid diverging more.


thayerandrews:
This all looks good to me.

PR #1103, opened 2014-12-05
Added CCEffectOutline

osinoleg:
Simple outline effect based on Laplacian filter/matrix. Does not use distancefields.


cocos2d/CCEffectOutline.h
```diff
@@ -0,0 +1,50 @@
...
+/** Color of the outline */
+@property (nonatomic, strong) CCColor* outlineColor;
+
+/** Outline pixel width of the outline */
+@property (nonatomic) int outlineWidth;
+
+/// -----------------------------------------------------------------------
+/// @name Initializing a CCEffectDFOutline object
+/// -----------------------------------------------------------------------
+
```
thayerandrews, line 31: It looks like CCEffectDFOutline should be CCEffectOutline.

cocos2d/CCEffectOutline.h
```diff
@@ -0,0 +1,50 @@
...
+/**
+ *  Initializes a CCEffectDFOutline.
+ *
+ *  @return The CCEffectDFOutline object.
+ */
```
thayerandrews, line 36: Same as above.

thayerandrews:
Add a couple comments about the doc comments but otherwise this looks good.

PR #1105, opened 2014-12-05
Fixed CCLabelTTF crash on iOS 6.0 devices.

osinoleg:
Addresses https://github.com/cocos2d/cocos2d-swift/issues/1098


hockeybro12:
Has this been fixed? Because I am using the latest version of Cocos2D using SpriteBuilder from the MAC App Store and it doesn't work still on ios 6.0 devices. The problem is when using custom founds with ios 6 and CCLabelTTF.

slembcke:
The fix is out in the 3.4 beta version, but the MAS store version is still 3.3.

PR #1106, opened 2014-12-05
Cleanup CCEffectImpl's initialization flow

thayerandrews:
This is yet another CCEffects cleanup change in preparation for Metal support. It's pretty big but for the most part reviewers can focus on the changes to CCEffect_Private.h and CCEffect.m. The remainder of the changes are sub-classes adapting to the super-class modifications.
- Simplify CCEffectImpl initialization flow. Prior to this change, an effect subclass called one of several initializers on CCEffectImpl of the form:

```
initWithArg:arg
initWithArg:arg blarg:blarg
initWithArg:arg blarg:blarg flarg:flarg
```

Which all chained to initWithArg: blarg: flarg:. The designated initializer then did the following:

```
if (!arg) [self buildArg];
if (!blarg) [self buildBlarg];
if (!flarg) [self buildFlarg];
```

And sub-classes could opt to override these builder methods. This change removes the overridable builder methods and, for the most part, requires the sub-classes to supply all components (functions, uniforms, varyings, etc) at the time that the super-class initializer is called.

This makes the super-class initializer call more verbose but it makes the overall flow more straightforward. It also addresses potential initialization order problems with overridden builder methods being called on the sub-class before the sub-class's initializer body could execute.
- The function, uniform, and varying arrays are now readonly to prevent them being modified after initialization.
- CCEffectImpl's self pointer is no longer captured by the render pass blocks. Instead the blocks capture the interface class that owns the CCEffectImpl which prevents capturing self before super-class initialization is complete (which is what we were doing before).
- Move some setup code that was duplicated between blur, bloom, and drop shadow into a shared function.


slembcke:
Sorry for the delay again. Didn't see anything that jumped out at me.

slembcke:
Just one comment after merging into the develop-v4 branch. The old Cocos color structs (ccColor4f, ccColor3b, etc) aren't deprecated in v3.x, but I've been encouraging people to use GLKVector4 instead.

I did deprecated the structs in the 4.0 branch. GLKVector types have a fairly complete API, and I want to standardize on using CCColor/GLKVector4 everywhere internally.

thayerandrews:
Thanks Scott. I'll also add cleanup of ccColor4f and friends to my v4 todo list for CCEffects.

slembcke:
I did make those changes for you already since they were compiler warnings. ;)

thayerandrews:
Got it. Thanks Scott.

PR #1109, opened 2014-12-06
CCFile

slembcke:
Initial implementation of the CCFile class. The basic functionality and tests were done so quickly that I decided to go ahead and implement gzipped and encrypted streams.

For encryption, I went with the XXTEA algorithm since it was really easy to use. It's not going to be foolproof anyway, so I didn't see much point spending more than a couple hours on it. I made a little command line utility to perform the encryption as well. It's pretty basic. It encrypts 4kb blocks of data at a time, where the first four bytes of each block are the total number of bytes in the block. Files will be rounded up to the next 4kb in size. ~~Right now the key I used for the tests is hardcoded. I figured eventually we'd probably add a SpriteBuilder setting to generate a key for a project and have the user copy paste it into their app delegate. Sort of like TexturePacker did `[CCFile setEncryptionKey:@"0123456789ABCDEF"];`~~ (This method has now been added.)


cocos2d/Support/CCFile.h
```diff
@@ -0,0 +1,119 @@
...
+
+ @since 4.0
+ */
+@interface CCFile : NSObject
+
+/**
+ Set the key used on encrypted files the app will be loading.
+ It's recommended to set this value only once in your app delegate.
+ 
+ TODO where would a user find or create such a key.
```
NickyWeber: I think this should not be our concern. There could be hints on how to improve security like obscuring the key when included within the app or ways to store one on a server.

UnitTests/CCFileUtilTests.m
```diff
@@ -10,7 +10,75 @@
...
+    NSURL *url = [[NSBundle mainBundle] URLForResource:name withExtension:nil];
+    XCTAssertNotNil(url, @"What happened to the file?");
+    
+    id plist = @{@"Foo": @"Bar"};
+    
+    CCFile *file = [[CCFile alloc] initWithName:name url:url contentScale:1.0];
+    XCTAssertEqualObjects(file.loadPlist, plist);
+    
+    // Could probably make a better test... but...
+    XCTAssertTrue(file.loadData.length > 0);
```
NickyWeber: Is this even needed? By comparing the plists contents loadData is tested indirectly.

NickyWeber:
Looks good. A few comments if I may:
- Is there a way to test a wrong key? What's happening then? 
- I guess the user can only determine errors during development time since most errors are written to the log. Any way to improve the feedback channel to recover from errors during runtime?
- Is it necessary to hide `initWithName:url:contentScale:`? What harm could be done if this is made public?
- What should the minimum set of convenience methods be to load data as a certain type?

LearnCocos2D:
Looks good, clean and simple. The only amendment I have is that it can be useful for debugging to see a CCFile's original "name", basically what it was initialized with. Unless that is what 'name' represents? If not it should probably belong to the private header, and maybe even being available only in debug builds.

slembcke:
@NickyWeber 
- No way to test for a bad key now. It will decode garbage which will probably be catastrophic for anything trying to read the data. I hadn't considered that.
- Largely true yes. I forward the steamStatus and streamError methods, but now that you point it out. That's not very correct. Do you think we should add NSError parameters  to the loadData and loadPlist methods?
- It's not necessary to hide the initializer, though I'm not sure if it's useful to expose it either. The `content scale` and `name` parameters are potentially strange for a file object created manually.
- Not sure I understand the last question. Do you mean the `loadData` and `loadPlist` methods? I added those since they are fairly raw formats. I think other types should be loaded using separate APIs (ex: images, particle definitions, CCB files, etc). I suppose we could add JSON too since that can decode nicely into plist types.

@LearnCocos2D 
Yes, the `name` property is supposed to be the original name passed to file utils before applying any name aliasing or suffixes: `[CCFileUtils fileNamed:@"SomeSprite.png"]`.

cocos2d/Support/CCFile.h
slembcke: By "where would a user find or create such a key" I mean documenting the teacrypt utility or documenting any encryption functionality we might add to SpriteBuilder.

UnitTests/CCFileUtilTests.m
slembcke: `loadPlist` calls the `openInputStream` method, but it doesn't call `loadData`. The reason why I took shortcuts with that test was because Xcode may or may not reencode the XML pilst file as a binary plist file, and it didn't seem important or fragile enough to create more testing files.

thayerandrews:
Regarding decoding with the wrong key, this post from Brent Simmons's
Vesper sync diaries comes to mind:

http://inessential.com/2014/04/17/vesper_sync_diary_14_keys

Most of it isn't exactly applicable but there is the second bullet which
discusses adding some magic tokens to the input data and making sure these
decrypt correctly at the expected location. Could we do something like that
for encrypted files?

slembcke:
I just pushed another commit with improved error handling and I've moved the encryption key method to the private header for now since there isn't a lot of consensus on it.

@NickyWeber, @thayerandrews 
Each compressed block stores its length in the first 32 bit word. The value has to be less than 4096, and I do already check it's length. That means there is a about a 1 out of 2^20 chance that the block length will be in the correct range if the key is incorrect. Right now it just throws an error that the stream is corrupt, but it should probably say that the stream is corrupt or the key is incorrect. Would that be "good enough"?

Quite seriously given the discussion, I might know the least about encryption about anybody commenting. I wouldn't feel bad if somebody wanted to rewrite it.

thayerandrews:
Has Collin chimed in on any of the encryption discussion? I don't see him
on this thread but I forget if he's spoken up elsewhere.

I ask because both he and Ian have a background in web security and they
would probably have some valuable insights on this topic.

-Thayer

NickyWeber:
@slembcke 
- I guess the NSError *\* is the standard pattern most people will recognize. I can't think of anything else reasonable in OBJC. 
- The encryption topic seems to be more complex, I'd say wait and don't touch this code wise until there is consensus. Martin is right, it's a nice selling proposition. Then the issue of bad keys can be addressed.
- Yes I was referring to loadData and loadPlist. I can think of many load methods :) Wondering if to move to a convenience category to keep code separated from all the many wonderful types. I guess anything else than loadData can be done with loadData and openInputStream.

slembcke:
A category with more load methods could be nice. I can't really think of anything more other than loadString or loadJSON perhaps. ;)

If there aren't any other non-encryption comments, I'll probably merge this tomorrow morning. Encryption or other extensions can be finalized later.

PR #1110, opened 2014-12-07
support for Mac OS X Retina displays

stuartcarnie:
Add retina support for Mac OS X. This is needed in order to merge in [SpriteBuilder #705](https://github.com/spritebuilder/SpriteBuilder/issues/705)

### DIFF
--- a/cocos2d/CCConfiguration.m
+++ b/cocos2d/CCConfiguration.m
@@ -233,9 +233,8 @@ static char * glExtensions;
 	}
 	
 #elif __CC_PLATFORM_MAC
-	
-	// XXX: Add here support for Mac Retina Display
-	return CCDeviceMac;
+    CCDirectorMac *dir = (CCDirectorMac *)[CCDirectorMac sharedDirector];
+    return dir.deviceContentScaleFactor == 1.0 ? CCDeviceMac : CCDeviceMacRetinaDisplay;
 	
 #endif // __CC_PLATFORM_MAC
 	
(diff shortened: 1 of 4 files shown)


cocojoe:
Anyone have any issues with this otherwise will merge, the Director changes are isolated to Mac Director. Always weary of `CCFileUtils` changes however shouldn't be an issue, @2x goes against our v3 Cocos2D conventions however makes sense for SB App.

stuartcarnie:
Agree on @cocojoe's **@2x** comment. The justification for supporting **@2x** when compiling for Mac was for apps like SpriteBuilder, that are developed primarily in Xcode and an application rather than a game – both of which benefit from the @2x suffix from a tooling perspective. Sketch 3 or Affinity Designer natively use the **@2x** suffix, so it removed some friction to knock this out relatively quickly.

cocojoe:
@slembcke Any thoughts on this? I would like to get this in for retina SB pull however throwing in @2x now seems a bit random however does make sense in an Mac App and seems unfair for @stuartcarnie to now rename files ;-)

stuartcarnie:
Another workflow benefit of the **@2x** suffix on a retina Mac is they are shown at the same size as their **@1x** counterparts in Finder and Xcode, but in higher resolution

slembcke:
I don't have a retina Mac to test on (iMac coming next week :D). I think it looks fine though.

There is already "-machd", which I honestly don't know if anybody uses. Using @2x on iOS actually breaks things since the image loading code still uses UIImage. Shouldn't make a difference on Mac though. We've committed to a complete overhaul of CCFileUtils for v4, so this will end up changing in the coming months.

PR #1111, opened 2014-12-09
CCAction unit tests, focused on reusing actions

andykorth:


cocojoe:
Sounds useful, I've heard a few issues of repeatability (v3), possibly when copying.

andykorth:
Oh, and for the curious, these are my findings regarding reuse and copying:

Copying an action:
- After copying, the action's target, elapsed time, and and "first tick" are all reset. A copied action is safe for using on a new node.
- This works for actions that have been completed or partially run. 
- This works for both for actions that have and have not been added to a node at the time of the copy.
- I didn't test other actions yet.

Reusing an action:
- Running the same action on two nodes doesn't throw any errors, but the action only affects the last node it was run on. (I think it should throw an assertion)
- If you call init again on an action that has been used, it will throw an assertion! Calling init again does not reset internal state variables. They are managed by CCActionManager by calling startWithTarget on them. We should definitely discourage people from calling init on anything a second time, since it's very strange thing to do anyway.
- There's no working way to reset an actions state, such as it's elapsed time.

PR #1112, opened 2014-12-09
CCBReader + CCLightNode - Support for light property animations

thayerandrews:
- Add CCBActionTweenColor, an internal action for tweening color properties.
- Handle CCLightNode's properties in CCActionManager's actionFromKeyframe0: andKeyframe1: ... method
- Implement copyWithZone: on CCActionTween to fix an exception when chaining tween sequences together. Without this we end up with an uninitialized tween object with a nil property key.

### DIFF
--- a/cocos2d-ui/CCBReader/CCAnimationManager.m
+++ b/cocos2d-ui/CCBReader/CCAnimationManager.m
@@ -230,18 +230,20 @@ static NSInteger ccbAnimationManagerID = 0;
         }
     } else if ([name isEqualToString:@"spriteFrame"]) {
         return [CCActionSpriteFrame actionWithSpriteFrame:kf1.value];
-    } else if ([name isEqualToString:@"intensity"] ||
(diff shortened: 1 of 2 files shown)


cocos2d-ui/CCBReader/CCAnimationManager.m
```diff
@@ -39,6 +39,19 @@
 // Unique Manager ID
 static NSInteger ccbAnimationManagerID = 0;
```
cocojoe, line 41: CCAnimationManager used to contain its own CCB Action methods however they were refactored into the CCAction family. `CCBActionTweenColor` would be more consistent as `CCActionTweenColor` and reside within `CCActionInterval`

thayerandrews, line 41: Thanks Martin. I was wondering about this part of this change. @vlidholt and I discussed making this particular class internal to CCBReader for now but it certainly could be made public instead if people would find it useful outside of this context.

PR #1120, opened 2014-12-12
Swift fixes for v3.4 (actions: sequence, spawn / physics: joint, shape)

LearnCocos2D:
Mainly exposing "initWith" in the header files so Swift bridging picks them up where the +(id) initializer isn't properly named.


cocojoe:
More Swift support the better.

PR #1124, opened 2014-12-17
CCViewportNode implementation

andykorth:
CCViewportNode provides users with an easy way to control a camera transform. Users can directly scroll or rotate the camera, or use CCActions to move the camera without having to move the entire scene in the opposite direction. 

In addition, CCViewportNode allows drawing to only part of a screen, and provides rendering and input culling. This can facilitate picture-in-picture displays (rear view mirrors, CCTV camera scenes, 3D character portrait stuff), and has some potential uses in head-to-head mobile games. 

For details and limitations, see: https://docs.google.com/document/d/1xFD98-jXi0k3B-Iyw4N3VT9U4c9U19qUA_3-c8yJ73M

Included in this pull request is an updated implementation of CCScrollView, which has gained the ability to cull it's content to a specific contentSize: http://kortham.net/temp/upshot_cSmDt12g.png Additional tests for this are included.

Not included: No render texture support yet. 
#### Name discussion:

As I've worked on it, I'm thinking we might want to rename this class to CCCamera. I think a beginner might look at something like "CCViewportNode" or "CCProjectionNode" and not really have any idea how that might help them. What's a viewport anyway? Isn't it something on a ship?? Although in [other engines](http://docs.unity3d.com/ScriptReference/Camera.html), the "camera" controls both what is drawn, and where on the screen it's drawn to, maybe we can come up with a better name. A few points relevant to the naming discussion:
- Moving the top level class, currently "CCViewportNode" results in moving destination viewport around on the screen.
- Moving what is currently called the myCCViewportNode.camera changes what is rendered into the viewport.
- Objects that are to be rendered inside the viewport are added to myCCViewportNode.contentNode

I'm looking forward to comments and reviews!


andykorth:
Oops, found some Mac issues with the CCScrollView/CCTableView stuff. be right back!

PR #1126, opened 2014-12-18
CCImage changes

andykorth:
Here is Scott's CCImage implementation, with some added unit tests.

The old proposal is found here: https://docs.google.com/document/d/1p5uhsyshRRasGStdEi0_sdBuGgbUSS7BypOji75O2pE


slembcke:
Note: A reimplementation of compressed tilemaps snuck into this pull request. The code for it was nearly identical to what embedded particles needed. I ended up doing a some of the refactoring from big v4 document as well. The commit probably should have gotten cherry picked into a separate pull request.

cocos2d/CCImage.h
```diff
@@ -0,0 +1,149 @@
...
+
+ @param pixelSize    Size of the image in pixels.
+ @param contentScale Content scale of the image.
+ @param pixelData    A pointer to raw, tightly packed, RGBA8 pixel data.
+
+ @return An image object that wraps the given pixel data.
+
+ @since 4.0
+ */
+-(instancetype)initWithPixelSize:(CGSize)pixelSize contentScale:(CGFloat)contentScale pixelData:(NSMutableData *)pixelData;
```
thayerandrews, line 84: Just curious, why NSMutableData and not NSData?

cocos2d/CCImage.h
```diff
@@ -0,0 +1,149 @@
...
+
+ @since 4.0
+ */
+@property(nonatomic, assign) CGSize contentSize;
+
+
+// TODO Other thoughts:
+// * Method to create a CGImage from the CCImage?
+// * Method to capture the screen as a CCImage?
+
```
thayerandrews, line 148: If a screen capture to CCImage method is needed, I'm not sure I'd put it this functionality in the CCImage class itself. That starts to make CCImage more complicated than just a container of pixels which seems counter its design.

thayerandrews:
Made a couple inline comments but this looks good.

andykorth:
Thanks for taking a look at it! Yeah, I agree about the screen capture stuff. Not sure where the best place to put it is, but I don't think that's super high on the priority list.

Thanks!

cocos2d/CCImage.h
slembcke, line 84: I waffled on that several times, but maybe it doesn't make sense. The original proposal used NSData.

@Birkemose's main suggestion was to make it possible to, for example, allow calculating normal maps at load time from a heightmap. I had been thinking about similar ideas about loading filters in general, so this fit in well. I changed it to mutable at some point so that it wouldn't require a copy to filter the image data.

On the other hand, very very few interesting image filters can be performed in place without a copy anyway. Also, since textures are uploaded on the rendering thread, this opens up the possibility for race conditions if the multi-threaded renderer is used. So... maybe mutable was a bad decision after all. It only saved one line of code for people applying filters to loaded image data, which is probably a very small number of people anyway.

cocos2d/CCImage.h
slembcke, line 148: Hmm. Since we are refactoring the director/view stuff. Maybe it would make sense to put an instance method on the view class that grabs the view's pixels as a CCImage. That might make more sense.

thayerandrews, line 148: My gut reaction upon reading this is that I prefer this idea. It seems like a better division of responsibilities if the owner of the view / screen knows how to create CCImage objects versus the CCImage class knowing how to access views and extract pixels from them. This keeps CCImage simpler and more platform independent which I like.

PR #1143, opened 2014-12-30
CCEffectLighting and CCEffectColorChannelOffset fixes

thayerandrews:
Fix a few small CCEffect issues:
- Fix for SB issue #1252. When CCEffectLighting is applied to a CCEffectNode it is unable to access the node's scene (and therefore the light collection) because of CCRenderTextureSprite's unique parentage / ownership. Handle this in a CCEffectUtils helper function and always use that to access the scene.
- Make all CCEffectColorChannelOffset methods that use GLKVector2 deprecated and remove all references to them in favor of the CGPoint versions.
- Change the intensity value of CCEffectBloom so internally it ranges from [0..2.5](the publicly visible value is still [0..1]). This results in more dramatic blooms when the intensity is turned all the way up.
- Make CCEffectOutline experimental for the 3.4 release.


slembcke:
I could swear this already got merged... There were related changes to this I had to merge into develop-v4. Merging now.

PR #1147, opened 2015-01-06
Update ObjectAL submodule.

rkachowski:
This review updates the ObjectAL submodule to point to a ref with an updated `.gitignore` file which suppresses false warnings about dirty submodules (intermediate build files being marked as untracked).


cocojoe:
My hero, was going to get round to this as I find it really annoying.

PR #1152, opened 2015-01-08
Updating develop with develop-temp for SB 1.4 release

KristenY:
Making sure everything from develop-temp makes it into develop.
I merged develop into develop-temp, and then made a pull request on develop to get this result.  Please check if these changes are ok to put into develop for the Cocos2D 3.4 release, which will be paired with SB 1.4 Mac App Store submission.


KristenY:
I'll do the same for the develop-temp version of SB once we're sure this one is okay.

Issue #1156, opened 2015-01-09
Separate Integration Tests from Unit Tests in cocos2d-tests.xcodeproj


rkachowski:
There should be no reason for a glcontext to be created whilst unit testing cocos2d in xcode. It should be possible to test components in isolation without requiring a fully configured cocos2d instance running in the background.

Currently, running the unit tests in `cocos2d-tests.xcodeproj` within develop @ `49ee7d` causes a cocos2d application to be launched in the simulator whilst the tests are run. This leads to complications with memory access, especially now in the case of the `CCPackage` test cases. These cases are currently ignored as they fail due to bad access errors caused by cocos2d trying to draw it's scene graph. 

This error can be encountered by setting `IGNORED_TESTS`  in `CCPackagesTestFixturesAndHelpers.h` to the value `0` and running the test suite in XCode.

slembcke:
For app target unit tests this is all standard Xcode behavior. It starts up the app normally, then hijacks the main event loop somehow. There is an option to run the test module in a clean process as well, but this has been an ignored issue for a while since many of the tests rely on singleton objects that rely on the (very particular) Cocos2D startup code. Most of the code involved has been around longer than anyone currently working on the project, so we've sort of worked around it instead.

This should finally be cleaned up in v4 though. \o/

PR #1157, opened 2015-01-09
Ignore async package tests.

rkachowski:
This PR is designed to address https://github.com/spritebuilder/SpriteBuilder/issues/1284 

The issue is that errors are being thrown in CCRenderBuffer during async tests that manipulate the render loop. The correct solution for this is to refactor integration tests (tests that require a running cocos scene) into a separate test target from the unit tests. This will require a non trivial amount of effort, and this patch is presented as an interim solution.

The issue #1156 is added to address the refactoring.

### DIFF
--- a/UnitTests/CCPackageDownloadManagerTests.m
+++ b/UnitTests/CCPackageDownloadManagerTests.m
@@ -12,6 +12,7 @@
 #import "CCPackage.h"
 #import "CCDirector.h"
 #import "AppDelegate.h"
+#import "CCPackagesTestFixturesAndHelpers.h"
 #import "CCUnitTestAssertions.h"
 #import "CCPackage_private.h"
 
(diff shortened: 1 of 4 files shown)


slembcke:
@andykorth made a similar fix for this in the develop-v4 branch. Part of the problem is some weird issues with how the director does it's startup. The fix in the v4 branch was to modify the director with some potentially breaking changes since it will all be replaced soon anyway. In the develop branch, I'm not sure of a better solution. I think people have just been running the tests more than once since they rarely crash twice in a row. Perhaps a passive aggressive version of testing. /me facepalm 

In the short term I don't think we will be making any more package or file utils changes before v4, so it's probably safe to just disable the tests. I'll wait for Nicky to comment since they are his tests.

rkachowski:
I guess the separation issues can take a back seat to the `CCDirector` replacement. I spoke with Nicky about disabling these tests before posting the PR and he was fine with it.

PR #1171, opened 2015-01-18
Use 'instancetype' for factory methods to improve Swift support

Ben-G:
When factory methods use `instancetype` as a return type Swift can infer convenience initializers correctly. When returning `id` this is not possible.

Currently a `CCSpriteFrame` needs to be initialized like this:

```
let spriteFrame = CCSpriteFrame.frameWithImageNamed(imageName) as CCSpriteFrame
```

When returning `instancetype` the initializer can be used like this:

```
let spriteFrame = CCSpriteFrame(imageNamed: imageName)
```

**Changing to `instancetype` will cause the first solution to no longer work - breaking backwards compatibility with existing Swift code.**

So I guess this should be changed for 4.0? It definitely would be nice to have the second more concise option available soon.

Note: I had refactored this in the following pull request, but had messed it up prior to the merge: https://github.com/cocos2d/cocos2d-swift/pull/467

### DIFF
--- a/cocos2d/CCAction.m
+++ b/cocos2d/CCAction.m
@@ -41,7 +41,7 @@
 
 @synthesize tag = _tag, target = _target, originalTarget = _originalTarget;
 
-+(id) action
++(instancetype) action
 {
 	return [[self alloc] init];
 }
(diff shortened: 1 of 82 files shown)


cocojoe:
Still early days for Swift projects and really like the improvement in code initializing. However as changing is breaking I think this should be for v4 although as v4 is on going not sure the best time to make a pull or if it should just be part of the v4 process @slembcke ?

andykorth:
Moving towards `instancetype` is good for obj-c projects too. As far as breaking existing swift code... the sooner we correct this, the better, I think. I'd say it's appropriate for v4 (which is now the develop branch).

slembcke:
I merged this into the 3.4 branch. We should do another audit of this later for v4 though. We might as well change all of the init methods to return instancetype at that point too.

PR #1173, opened 2015-01-21
Refactor/CCFileUtils-V4

NickyWeber:
New implementation of CCFileUtils with new behaviour.
- Simplified API
- Simple resolution tagging: -1x, -2x, -4x
- Localization support for any file

_There's a new class CCFileUtilsV2 for now. Needs to be renamed if old one can be deleted._


cocos2d/Support/CCFile_Private.h
```diff
@@ -55,6 +55,13 @@ typedef NSInputStream *(^CCStreamedImageSourceStreamBlock)();
...
 
 
+@interface CCFile()
+
+@property (nonatomic, assign, readwrite) BOOL useUIScale;
+
+@end
+
+
 @interface CCFile(Private)
```
andykorth, line 65: Maybe this named category should have the name removed, and useUIScale could just go below?

UnitTests/CCUnitTestAssertions.h
```diff
@@ -1,19 +0,0 @@
-#define CCAssertEqualStrings(a1, a2, format...) XCTAssertEqualObjects(a1, a2, format)
```
andykorth, line 1: +1 for removal of CCUnitTestAssertions.h
Yay!

cocos2d/Support/CCFileUtilsV2.h
```diff
@@ -0,0 +1,140 @@
...
+ @since 4.0
+ */
+@property (nonatomic, assign) NSUInteger untaggedContentScale;
+
+/**
+ The device's content scale.
+ 
+ @since 4.0
+ */
+@property (nonatomic, assign) NSUInteger deviceContentScale;
```
slembcke, line 111: We do need to find a new place for the device's global content scale factor, but CCFileUtils is not really the right place since it's used for more than just handling files.

You don't need to change this right now though.

cocos2d/Support/CCFileUtilsV2.h
```diff
@@ -0,0 +1,140 @@
...
+ */
+@property (nonatomic, copy) NSArray *searchPaths;
+
+/**
+ A database that can be queried for metadata and filepaths of an asset.
+ Refer to the CCFileUtilsDatabaseProtocol for more details about the interface required.
+ 
+ @since 4.0
+ */
+@property (nonatomic, strong) id <CCFileUtilsDatabaseProtocol> database;
```
slembcke, line 96: Should this be readonly?

slembcke, line 96: Another possible problem. `fileNamed:error:` will apply search path rules and return a specific file.

To get the metadata for that file, I would have to do something like this:

```
[[CCFileUtilsV2 sharedFileUtils].database metaDataForFileNamed:filename inSearchPath:searchPath];
```

How do you get the correct search path without trying them all in order? Does it make sense instead to move the `metaDataForFileNamed:inSearchPath:` method to file utils and make the search path argument optional? The CCFileUtilsDatabaseProtocol could be made private as well.

cocos2d/CCFileUtilsConstants.m
```diff
@@ -0,0 +1,30 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+
+#import <Foundation/Foundation.h>
+
+NSUInteger const ERROR_FILEUTILS_NO_SEARCH_PATHS = 20000;
```
slembcke: Any reason to put these in their own .h/.m file instead of the main file utils header?

cocos2d/Support/CCFileUtilsV2.m
```diff
@@ -0,0 +1,373 @@
...
+
+    TraceLog(@"* Searching for content scale variants: %@", searchVariants);
+
+    for (CCFileUtilsV2SearchData *fileLocatorSearchData in searchVariants)
+    {
+        NSURL *fileURL = [NSURL fileURLWithPath:[searchPath stringByAppendingPathComponent:fileLocatorSearchData.filename]];
+
+        BOOL isDirectory = NO;
+        NSFileManager *fileManager = [NSFileManager defaultManager];
+        if ([fileManager fileExistsAtPath:fileURL.path isDirectory:&isDirectory] && !isDirectory)
```
slembcke, line 294: Need to add a check for a local URL.

Until now, I hadn't considered that you can't just query for file existence on remote servers. We would have to make the request and check for errors. This could make loading files from HTTP _ridiculously_ slow if it has to make a few failing attempts before it finds the right file.

My main reason for wanting to support remote servers was to support reloading assets at runtime on a device. The performance wouldn't really be a problem there. I guess people would figure out the performance implications of remote file pretty quickly.

cocos2d/Support/CCFile_Private.h
NickyWeber, line 65: That was the first thing I did but It didn't work so I moved it to a class extension.
Most of the CCFileUtilsV2Tests fail with `-[CCFile setUseUIScale:]: unrecognized selector...` when I move the overridden property to the private category.

cocos2d/Support/CCFileUtilsV2.h
NickyWeber, line 111: Ok, what about this new CCDeviceInfo class?

cocos2d/Support/CCFileUtilsV2.h
NickyWeber, line 96: Ah I see, this is more a design than a usage decision: Idea is to inject dependencies from outside of the class. Some class will configure the database for fileutils. Donald was working already on something central. I'd add some changed to the cocos2dConfig.plist file generated by SB to include the database.

So it's file utils asking the database configured for metadata instead of you querying the db. 

Without that database there would be just no support for aliasing, localization and UIScale.

cocos2d/CCFileUtilsConstants.m
NickyWeber: I guess it's me moving dependencies into more remote places. Will move it to the utils header.

cocos2d/Support/CCFileUtilsV2.m
NickyWeber, line 294: Please elaborate on how this check should work. Right now the search is finding a file by going through that list of possibly available files. Every resolution variant is tested for existence and then returned if it does. 

To skip this test(whether local or remote) for a searchPath that is not local(Don't know yet but I guess a test for a path beginning with a / should suffice for locality) would mean to return the first entry in that search list of resolution variants. That would be guess work I think.

UnitTests/CCUnitTestAssertions.h
NickyWeber, line 1: Was my fault in the first place, didn't recognize for a while that EqualObjects is doing string comparison just fine :)

cocos2d/Support/CCFile_Private.h
slembcke, line 65: Oh. That's a really bizarre auto-property synthesis rule. If you put it in a named extension, you have to use the `@synthesize` keyword in the implementation. I'm honestly not sure why Xcode automatically synthesizes properties in anonymous categories. Seems problematic to me.

cocos2d/Support/CCFileUtilsV2.h
slembcke, line 111: Yeah, I'm thinking it makes sense to either go there or into the app delegate since that is where the initial configuration actually happens. (I still haven't reviewed Donald's changes for that...)

cocos2d/Support/CCFileUtilsV2.m
slembcke, line 294: Hmm. Actually, it looks like [NSURL checkResourceIsReachableAndReturnError:] might do exactly what we need. It shouldn't matter if the URL is local or remote.

What I meant originally was that you can't use NSFileManager to check if a file exists on an http:// url, only a file:// one. If the URL method above does what we need, then it should be an easy fix. If the solution is too much more work, we should probably move the remote URL feature to a later release.

cocos2d/Support/CCFile_Private.h
NickyWeber, line 65: So move and add @synthesize or leave it as it is?

cocos2d/Support/CCFileUtilsV2.h
NickyWeber, line 111: Please go ahead I think the appdelegate are really slimmed down in his PR.

cocos2d/Support/CCFileUtilsV2.m
NickyWeber, line 294: Ok I guess we should keep it simple as for and leave it. 

About checkResourceIsReachableAndReturnError:

> This method is currently applicable only to URLs for file system resources. For other URL types, this method always returns NO.

Question is what to do if a requested image and, say the second available search path is an http:// url so what to do in the algorithm? Check for localitly: No, ok don't use the filemanager. Without actually requesting the files on a remote host(can do a HEAD request to keep overhead small and simultaneously request all possible variants) it's not possible to return something. CCFileUtils then needs to be usable in a non blocking way.

slembcke, line 294: Yeah, I guess that's what I mean. Checking the existence of files over HTTP isn't trivial, and doing it synchronously is a _really_ bad idea.

Definitely an oversight. Let's drop remote URL features for now. A couple of the CCFile features (streams) don't work yet either. We can always revisit the idea in a minor version.

Don't worry about changing anything right now.

NickyWeber:
I removed the constants file and moved the error constants to the ccfileutilsv2 header.

slembcke:
Thought a bit more about fileNamed/imageNamed. Internally, Cocos is going to need to use the imageNamed method to load a few things that are not images. Spritesheet plists, bitmap font definitions, possibly TMX files, and possibly particle definition plists. (The latter two are caused by legacy issues from 1.x that _really_ should be fixed)

Is it weird to locate image related files using imageNamed if it's internal to Cocos2D? Are there any files a user might want to load with the image resolution search in their own code?

NickyWeber:
@slembcke I can actually live with it. We could also remove `imageNamed` and add an option or rename `imageNamed` to `fileNamedWithResolutionSearch` or something alike. It's just a convenience to separate both cases, let's find a good name for it, I'd rather not add an option.

slembcke:
I'm okay with leaving it for now too. `fileNamedWithResolutionSearch` is a bit verbose, but it's very clear. I think I would prefer that over an options parameter too.

andykorth:
I discussed this with Scott. I thought it was a little weird for something starting with `imageNamed` to not return a CCImage, but it's more representative of it's actual use. I'm also okay with the longer `fileNamedWithResolutionSearch` method name. Either is good.

slembcke:
Your unit tests were fairly complete, so I decided to try out the API by giving the integration a test run. Overall it went pretty well. I removed the old CCFileUtils and fixed all of the errors. Half of the errors were in the unit tests (which I didn't try to fix).

Tada!


My only feedback is that using imageNamed seemed a little weird for non-images. We should think about renaming it to the more verbose version at a later point. Also, there were a number of occasions where a file was directly loaded into a UTF8 string. So it would probably be useful to add a method for that onto CCFile.

I didn't rename all of the files, so only some of the tests worked properly. There were some resolution/size issues, but I'm relatively certain they were in my hacked together integration and not CCFileUtilsV2.

I pushed my commits here if it's useful to see, but it might be easier to do them over again than to cherry pick them.
https://github.com/slembcke/cocos2d-iphone/tree/refactor/ccfileutils

andykorth:
Nice work with these changes! I am excited to use them!

slembcke:
Oops. I was going to make one more comment about names before merging it. I guess it doesn't really matter. Since nothing references it yet.

I'm okay with either CCFileUtils or CCFileLocator as the class's final name. "Locator" is more descriptive I guess. @NickyWeber, I'll leave the decision up to you.

NickyWeber:
@slembcke Ok I created a PR for some minor changes. I renamed the class to locator to make a clean cut. Tests are included in test target again, must have been removed by the director merge. 

How do we proceed? Should I get your integration branch and work on it further or will you make the transition to the new locator?

PR #1174, opened 2015-01-22
CCDirector Refactoring and un-singletoning

andykorth:
CCDirector Refactoring:

The primary goal behind this pull request is to un-singleton the CCDirector. This allows for better scene control, better testability, and clearer separation of platform specific code. CCViews own the reference to the CCDirector, and CCDirectors manage and own individual CCScenes.

Most of the credit goes to Oleg and Scott. Thanks!

Sorry, this is a huge pull request. I was initially breaking it down into smaller commits, but found each one depended on the previous. This pull request is going out now to prevent even larger ones from happening in the future. While this refactoring is not entirely complete, the project is currently working and usable. iOS unit tests, iOS testbed, and Mac has been recently tested. Oleg is working on Android.

Highlights of changes
- CCDirector is no longer a singleton, but the current director is accessible, similar to how CCRenderer is accessible.
- CCView is a common protocol to each platforms view implementation.
- CCScenes have a clearer relationship with directors.
- CCActions were refactored to reflect the possibility that they were added to nodes that haven't yet been added to a scene
  *\* CCActionManager is removed and merged into CCScheduler. There was a lot of common functionality there without any common code.
  *\* Merging these classes unified pausing behavior. 
  *\* Pretty big performance improvements both for actions and scheduled timers. (calling update on nodes, etc)
- Refactor scene changing methods. Use presentScene exclusively to change scenes, whether it's the first scene or not.

Areas of continued development:
- Running CCActions in the fixed update loop.
- Queuing scheduled timers on nodes that haven't been added to a scene
- Fixing up test like the package manager tests (testability without a view, etc)
- ContentScale moving out of director.
- Stats label
- Application lifecycle changes w/ director.

Ultimately, there are too many people working on highly related code for these changes not to live in a commonly accessible place. Even if these changes aren't super stable, the alternative is people writing large amount of code independently and having huge merge conflicts at the end. I'm not sure what exactly the stability expectations are of the "develop" branch, but that would come at the cost of regular integration and conflicts.


cocos2d/Platforms/Mac/CCViewMacGL.m
```diff
@@ -0,0 +1,296 @@
...
+{
+    CCLOGINFO(@"cocos2d: deallocing %@", self);
+}
+
+#pragma mark CCGLView - Mouse Delegate
+
+- (void)mouseDown:(NSEvent *)theEvent
+{
+    // dispatch mouse to responder manager
+    [CCDirector bindDirector:_director];
```
rkachowski: What is the benefit here? It looks like we are adding 2 lines of boilerplate overhead to every `CCDirector` call

andykorth: Yeah, unfortunately the bindDirector calls are pretty verbose in this section. The reason for this is that each one of these events is an entry point into cocos2d code from various NS input handling places. Further down this stack trace, we will have user code responding to mouseDown events. Inside that user code, the user might be looking for the `[CCDirector currentDirector]` for that thread, and it needs to be bound for them. 

I'm open to other suggestions.

It could be moved into the scheduler, but I think that's just pushing the binds into bigger methods where it's not as easy to see what's going on. (although looking at it, there would be a few less calls).

And it's important to unbind the director to ensure correctness. It makes it much easier to find a mistake like a missing bind in the engine, rather than leaving the 'usually correct' director bound, causing infrequent race-condition like issues when the user's setting up a new scene or something like that.

Thank you for reviewing!

rkachowski: As far as the duplication goes, at the least I would create some kind of method to encapsulate the binding/unbinding of all this. e.g. `[foo withDirector:_director perform:^(){[_director.responderManager mouseDown:theEvent];}]` - also the (duplicated) comments here are quite superfluous, as the code is rather explicit about what it is doing.

On a larger architectural scale, it seems pretty inefficient to forward every event to a secondary object with the exact same interface. There's no benefit of platform abstraction here. Why not have a single `handleUIEvent:` method on the responderManager, which will invoke different logic depending on the event type? as it stands there is a 1:1 mapping between the API of these classes.

Is there a situation where different threads would have different directors? The continuous binding and unbinding feels like a rather major anti-pattern to introduce into the codebase.

andykorth: Good ideas! The extra mirrored methods in CCScheduler are definitely superfluous. I'll update the request. The CCViewMacGL has way more events than iOS, they can definitely be sent to a much smaller number of already existing methods. It could be converted to mouse, scroll wheel, and key events pretty trivially, and could be convered to a single handleUIEvent method too (although organizationally that might be a bit of a big method).

Yeah, I should have removed those nonsense comments when I was touching it. CCViewMacGL is a renamed CCGLView, which is pretty old. Will cleanup.

There are situations where the same thread could have different directors (and different views). This is potentially addressing situations where a game is running with two CCViews, each one would have different directors. I'm guessing we can't make any assumptions over which thread is calling these input methods. Maybe @slembcke can comment more on this. It also improves testability, when we can create fresh CCDirectors for each test.  Any other ideas for improving this? The only one that really comes to mind is "bring back the singleton" or do a better job of encapsulating the bindDirector calls (which I am on!).

slembcke: One of the biggest problems with the director being a singleton is that you can only have one cocos2D view in an application, and it makes the startup process _really_ convoluted. (as I'm sure you've discovered) Since it's not accessible as a global anymore, it either needs to be passed around as an argument or set as a context value. Passing the director as an argument to all update methods, events, user callbacks, etc is also a little crazy and a _lot_ more work for both us and users. Setting it as a context value lets it work in a more or less backwards compatible way for the majority of users that will only ever use a single view in their application.

We don't really have a lot of choice in forwarding the methods I don't think. Either every view type (currently 4 of them) needs to implement it's own event dispatching, or we need to forward the events to a common object that is shared by all of the platforms. Efficiency isn't really a problem. Events happen much less often than the rendering loop, so a few extra method calls aren't going to hurt the performance. It's Basically the same reason why we decided not to remove the director after thinking about it more. The code that is currently in the director would just need to be duplicated by 4 different view classes. On the other hand, we now have the opportunity to merge the 3 director classes into one since there isn't very much x-platform code in it anymore that couldn't move to the view classes.

Since we can't really break the dependence on having a director, or at the very least, knowing the view that a certain scene is running as a part of, we don't really have any other ideas of what to do.

Also, I suppose it's important to note that the mouse\* events are defined on NSResponder so we can't really change how that works.

andykorth: Yeah NSResponder could have been shorter.

Opinions on these selectors? Mash them together? Not sure if I should change something that will break user code, but if there's ever a chance for an API change, it's now. Plus I bet the the mac version doesn't get a lot of use- only a few people noticed there was no keyboard support in v3.
https://github.com/andykorth/cocos2d-iphone/blob/CCDirector/MergeSchedulerAndActionManager/cocos2d/CCResponderManager.m#L549

andykorth:
Thoughts on these changes? Removed all the redirection methods from CCRespondermanager here: https://github.com/andykorth/cocos2d-iphone/commit/45d0184a8e3a5db6d8533d8ccc4db29aa7fb3b97#diff-9359b51e5d023b91282b75346faf53e6L607

Made the CCViewMacGL ones shorter, and call the right methods directly. 
https://github.com/andykorth/cocos2d-iphone/commit/45d0184a8e3a5db6d8533d8ccc4db29aa7fb3b97#diff-df9801ff9df3d2733cc1388166fd851cL175

Moved the bindDirector calls into the actual responder methods, the main plus being that there's fewer of them there. I think the event handling methods in CCResponderManager could be integrated a bit more, but that will have to wait til after the Global Game Jam!

NickyWeber:
Is there actually demand to make games with more than one view?

cocos2d/Platforms/Mac/CCViewMacGL.m
rkachowski: wrt the discussion: in what situation would you wish to have more than one CCView in an application? The concept of multiple directors makes sense in this situation if you have more than one view and wish to display multiple scenes simultaneously, but I can't seem to envision a use case that can't be achieved with render targets.

As far as testability goes, I feel this issue is due to the dependencies of discrete objects within cocos2d on a director rather than the fact that it exists as a singleton. I'd like to point out that I'm in no way attached to the implementation of the director as a singleton.

I'm totally for moving the platform specific code out of the director instances, this would really help differentiate between platform / GL initialization setup and what is relevant for cocos configuration. 

I've started with a pull request to make this clearer with regards to https://github.com/spritebuilder/SpriteBuilder/issues/912, but it has significant crossover with your changes here and I don't want to complicate your efforts (esp with the impending v4 deadline) so I will wait for this to go in.

cocos2d/CCActionManager_Private.h
```diff
@@ -24,5 +24,5 @@
 #import "CCActionManager.h"
 
 @interface CCActionManager ()
--(void)migrateActions:(id)target from:(CCActionManager*)oldManager;
+// TODO eliminate file
```
rkachowski: why not remove it now?

cocos2d/CCDirector.h
```diff
@@ -66,13 +66,13 @@
 /** Stops the animation. Nothing will be drawn. The main loop won't be triggered anymore.
  If you want to pause your animation call [pause] instead.
  */
--(void) stopAnimation;
+-(void) stopRunLoop;
```
rkachowski, line 70: The comment above will need to be changed to reflect the functionality and so the documentation doesn't conflate "animation" and "runloop"

cocos2d/CCDirector.h
```diff
@@ -66,13 +66,13 @@
...
 
 /** The main loop is triggered again.
- Call this function only if [stopAnimation] was called earlier
- @warning Don't call this function to start the main loop. To run the main loop call runWithScene
+ Call this function only if [stopRunLoop] was called earlier
+ @warning Don't call this function to start the main loop. To run the main loop call presentScene
  */
--(void) startAnimation;
+-(void) startRunLoop;
```
rkachowski, line 74: I would recommend renaming this, as having a method `startRunLoop` along with the attached warning that states "don't call this to start the loop" is pretty counter intuitive

cocos2d/CCDirector.h
```diff
@@ -489,15 +472,15 @@ typedef NS_ENUM(NSUInteger, CCDirectorProjection) {
...
  
- @see startAnimation
+ @see startRunLoop
  */
--(void) stopAnimation;
+-(void) stopRunLoop;
 
 /** Begins drawing the screen. Scheduled timers and actions will run.
  
  @warning Don't call this function to start the main loop. To run the main loop call presentScene:
```
rkachowski: as above

cocos2d/CCDirector.m
```diff
@@ -160,30 +153,10 @@ - (id) init
 		// FPS
 		_displayStats = NO;
 		_totalFrames = _frames = 0;
-
-		// paused ?
```
rkachowski, line 164: removing these kind of comments is awesome :+1:

cocos2d/CCLabelTTF.m
```diff
@@ -479,7 +477,7 @@ - (CCTexture*) createTextureWithAttributedString:(NSAttributedString*)attributed
     
     CGSize originalDimensions = _dimensions;
   
-    CGFloat scale = [CCDirector sharedDirector].contentScaleFactor;
+    CGFloat scale = __ccContentScaleFactor;
```
rkachowski, line 480: i would change this to  `CGFloat scale = self.director.contentScaleFactor` as you have on line 706

cocos2d/CCNode_Private.h
```diff
@@ -81,4 +75,22 @@ CGPointApplyGLKMatrix4(CGPoint p, GLKMatrix4 m){
...
+
+/**
+ In certain special situations, you may wish to designate a node's parent without adding that node to the list
+ of children. In particular this can be useful when a node references another node in an atypical non-child
+ way, such as how the the CCClipNode tracks the stencil. The stencil is kept outside of the normal heirarchy,
+ but still needs a parent to function in a scene.
+ 
+ @since v4.0
+ */
+-(void)setRawParent:(CCNode *)parent;
```
rkachowski, line 87: ~~The comment quite clearly explains the use for this, but it feels a bit awkward to add this property to every instance + subclass of `CCNode`, just to support the stencil / clipping node functionality.~~

forget that, i just noticed this is a private header :)

cocos2d/CCResponderManager.m
```diff
@@ -695,29 +629,49 @@ - (void)scrollWheel:(NSEvent *)theEvent
...
- */
+#pragma mark - Mac keyboard handling -
 
-- (void)mouseMoved:(NSEvent *)theEvent
+- (void)keyDown:(NSEvent *)theEvent
 {
-    
-}
+    if (!_enabled) return;
+    //  [[CCDirector currentDirector].view.window makeFirstResponder:[CCDirector currentDirector].view];
```
rkachowski, line 641: remove?

cocos2d/CCResponderManager.m
```diff
@@ -695,29 +629,49 @@ - (void)scrollWheel:(NSEvent *)theEvent
...
+    }
+    [CCDirector bindDirector:nil];
     
 }
 
-- (void)mouseExited:(NSEvent *)theEvent
+- (void)keyUp:(NSEvent *)theEvent
 {
+    if (!_enabled) return;
+    //  [[CCDirector currentDirector].view.window makeFirstResponder:[CCDirector currentDirector].view];
```
rkachowski, line 658: also

cocos2d/CCScene+Private.h
```diff
@@ -1,9 +1,7 @@
 /*
```
rkachowski, line 1: git has really screwed up and thinks this is a rename operation..

cocos2d/CCScheduler.m
```diff
@@ -71,32 +76,190 @@ @interface CCTimer (Private)
...
-@implementation CCScheduledTarget {
-	__unsafe_unretained NSObject<CCSchedulerTarget> *_target;
-	CCTimer *_timers;
+//MARK: Copy on Write Arrays
+
+// This is an NSMutableArray-like class that implements copy-on-write to allow modifying an array while iterating it.
+// This is for performance reasons so that we don't need to copy large arrays every frame to iterate them safely.
+// Instead, the copy is only performed when the array is actually changed.
+
+@interface CopyOnWriteArray : NSObject<NSFastEnumeration> @end
```
rkachowski, line 94: would you consider moving this into its own file?

cocos2d/CCScheduler.m
```diff
@@ -137,19 +300,20 @@ -(void)setPaused:(BOOL)paused
...
 @end
 
 
-@implementation NSNull(CCSchedulerTarget)
+@implementation NSNull(CCSchedulableTarget)
 -(NSInteger)priority {return NSIntegerMax;}
 @end
 
 
+//MARK: CCTimer
```
rkachowski: is this some kind of IDE specific instruction, or do you mean `#pragma mark` ?

cocos2d/CCScheduler.m
```diff
@@ -517,7 +693,107 @@ -(void)update:(CCTime)dt
 	[self updateTo:_currentTime + clampedDelta];
 	
 	InvokeMethods(_updates, @selector(update:), clampedDelta);
+    
+    #warning TODO: also invoke update actions:
```
rkachowski: the line below invokes `updateActions` or does this warning refer to something else?

cocos2d/CCTextureCache.m
```diff
@@ -82,9 +82,13 @@ -(id) init
 		// Skip the GL context sharegroup code for Metal.
 		if([CCDeviceInfo sharedDeviceInfo].graphicsAPI == CCGraphicsAPIMetal) return self;
 		
-#if !__CC_PLATFORM_ANDROID
-		CCGLView *view = (CCGLView*)[[CCDirector sharedDirector] view];
+
+#if __CC_PLATFORM_IOS
+		CCViewiOSGL *view = (CCViewiOSGL*)[[CCDirector currentDirector] view];
 		NSAssert(view, @"Do not initialize the TextureCache before the Director");
```
rkachowski: you can move this comment outside of the `#if` clause

cocos2d/CCTextureCache.m
```diff
@@ -82,9 +82,13 @@ -(id) init
...
+#elif __CC_PLATFORM_MAC
+        CCViewMacGL *view = (CCViewMacGL*)[[CCDirector currentDirector] view];
+        NSAssert(view, @"Do not initialize the TextureCache before the Director");
```
rkachowski: same

cocos2d/CCTextureCache.m
```diff
@@ -238,7 +242,7 @@ -(void) addImageAsync: (NSString*)path target:(id)target selector:(SEL)selector
 //            [EAGLContext setCurrentContext:nil];
```
rkachowski: i would gratuitously kill all these comments instead of just updating the director referencing - the previous code exists in version control should we need to reference it

cocos2d/Platforms/Android/CCGLView.h
```diff
@@ -41,9 +41,8 @@ enum CCAndroidScreenMode {
     CCScreenScaledAspectFitEmulationMode
 };
 
-BRIDGE_CLASS("com.apportable.GLView")
-@interface CCGLView : GLView <CCDirectorView>
-
+BRIDGE_CLASS("org.cocos2d.CCGLView")
```
rkachowski, line 44: is this correct? afaik there should be no `.java` files in cocos2d, and instead bridged from `GLActivityKit` - there is no `CCGLView.java` class to bridge with

cocos2d/Platforms/Mac/CCViewMacGL.m
```diff
@@ -0,0 +1,232 @@
...
+    [[self openGLContext] makeCurrentContext];
+    
+    // Synchronize buffer swaps with vertical refresh rate
+    GLint swapInt = 1;
+    [[self openGLContext] setValues:&swapInt forParameter:NSOpenGLCPSwapInterval];
+}
+
+- (NSUInteger) depthFormat
+{
+    return 24;
```
rkachowski, line 46: i would create a named constant to give context to this number (i.e. bitdepth)

cocos2d/Platforms/Mac/CCViewMacGL.m
```diff
@@ -0,0 +1,232 @@
...
+    
+    CCDirector *director = _director;
+    [director reshapeProjection: NSSizeToCGSize(rect.size) ];
+    
+    // avoid flicker
+    // Only draw if there is something to draw, otherwise it actually creates a flicker of the current glClearColor
+    if(director.runningScene){
+        [director mainLoopBody];
+    }
+    //	[self setNeedsDisplay:YES];
```
rkachowski: remove?

rkachowski:
My main concern is that the `currentDirector` calls rely on an implicit state existing that isn't directly clear to the end user, where operations will tend to fail if bind/unbind isn't called around calls when the user attempts to interact with the director.

cocos2d/CCDirector.h
andykorth, line 74: Ah. I think that warning means the user should never call this method. Honestly it should be in a private header. 

It is the method that starts the main loop, but the user is supposed to present a scene, which internally calls this.

The only situations when the user may want to start and stop the run loop is- when a mobile app is suspended (as a lifecycle event), or if they are moving in and out of a cocos2d game within something like a UIKit-based app. 

"startAnimation" was a very confusing name, I thought it had something to do with CCAnimations, etc. I'm open to other suggestions, other than "startRunLoop". But is this problem fixed if I just correct the comments and move these to a private header?

rkachowski, line 74: Yeah a private header would hide this from a user and fixing the comment would make more sense.

cocos2d/CCScene+Private.h
andykorth, line 1: Speaking of which, CCScene+Private.h should be CCScene_Private.h to be consistent. Will fix.

Apparently the only common lines is the header and the "@end". Haha

cocos2d/CCScheduler.m
andykorth, line 94: Hmm, maybe. It's a completely internal class though, only used within CCScheduler.m

Just as a matter of length of the file, CCTimer could be moved out. It is used externally. (and in user code)

cocos2d/CCScheduler.m
andykorth: In xcode, `#pragma mark` and `//MARK` do the same thing. Consistency would be good. There's 692 `#pragma mark`s and 120 `//MARK:`s. I'll change the comments to pragmas for consistency, and I'll skip the ones in chipmunk (which looks like most of them)

cocos2d/CCScheduler.m
andykorth: This is one of the unfinished items. Specifically fixed update actions.

andykorth:
@NickyWeber Good question. While in theory it can be done, I think real benefits occur in situations like: UIKit apps that run some cocos2d content inside a mostly Cocoa app, which they might navigate away from and back to. Support for multiple views is mostly just a side benefit of un-singletoning the CCDirector.

cocos2d/CCScheduler.m
slembcke: Actually I'd do it the other way around. The pragma version came first, then the comment version. I think the reason was because pragmas are supposed to be for compilers, not IDEs. You are actually supposed to put ifdefs around pragmas since they are compiler specific and will cause an errors in other compilers. (Not that it's a problem here)

cocos2d/Platforms/Android/CCGLView.h
slembcke, line 44: Not sure if those changes were merged into this branch. (There are too many to keep track of right now...)

andykorth, line 44: I'm going to CC @osinoleg on this. I haven't looked into any of his android changes here. I know he's working on it.

slembcke:
Right. It's actually quite difficult, and _extremely_ poorly documented on how to use Cocos2D within a larger application, even if you only want one Cocos view at a time. This does come up pretty often.

One of the benefits of SpriteKit is that it works in a regular UIKit/AppKit way. A SKView is just another view that you can create and destroy as you please. I can't count the number of times I've read comments about how this is a killer SK feature. Cocos2D was designed to control the entire app, and doesn't like to share.

Think about the way Cocos2D integrates with SB. You have a single document view that is open for the entire duration of the app. Lots of shared state when you close one project and open another. No possible support for having two projects open at once. SB is designed around these limitations.

@rkachowski 
We should be able to cover all of the entry points pretty easily for anybody using Cocos in a traditional way (one director/view for the entire app's lifecycle). We'll need to document how currentDirector works for the people that want to do more, but it's not more complicated than other contexts (GL, CoreAnimation, etc). I agree that contexts are annoying, and prefer APIs without them, but I don't have an alternative that makes it possible to extend the functionality and still remain backwards compatible.

cocos2d/CCScheduler.m
andykorth: Hm, I initially agreed, but if I'm going to find/replace them all, that should be a separate pull request. I don't want to delay this or make the pull request even bigger.

PR #1175, opened 2015-01-23
A little helper for the console to dump a node hierarchy.

NickyWeber:
Kudos to Donald: http://donaldhutchison.info/articles/ccnode-recursive-description/
As in UIKit's recursiveDescription and AppKits _subtreeDescription.


slembcke:
Nice. I've written a similar snippet of code a few times now when debugging various projects. I never bothered to copy it into the actual source though. -_-

PR #1176, opened 2015-01-23
Raised max touches to 11 to avoid iPad crashes

boblemarin:
A fix for bug 1166 (11 touches make ipad app crash)
https://github.com/cocos2d/cocos2d-swift/issues/1166


slembcke:
Hmm. I suppose we should audit the code and see if it's possible to remove the hard limit. I'll merge this for now though.

Issue #1122, opened 2014-12-16
CCTextField textField property is always nil


TimPelgrim:
I looked into the code and it seems that the property relies on the "nativeTextField" property of the CCPlatformTextField which in the current version I am using is not set in CCPlatformTextFieldIOS.

jonnyijapan:
This used to work in like 3.2, but stopped working along the way... this really needs fixing!

RobBoluga:
It's now been merged into V3.4

jonnyijapan:
Not entirely sure but I tried using your fix (thanks!) but it seemed like it had some problems with hiding/unhiding the uitextfield when CCtextfield was hidden/unhidden. I still have some more testing to do... busy with other things as usual... I'll just say the error can be on my side... Well this all used to work back last summer with whatever version was out then.

RobBoluga:
This is because in CCTextField the `update:` is constantly setting the visibility of the UITextField based on the visibility of itself and it's parent nodes. So it's a separate issue.
Why do you want to set `isHidden` on the UITextField rather than `visible` on the CCTextField?

jonnyijapan:
Anyway, I tried this out on a quick test project and it seems to work. My real project issue was probably caused by something else.

richardgroves:
@RobBoluga I've had a case where I wanted to hide just the UITextField and not the whole node - basically when covering up a CCTextField with another dialog - it looked better to just remove the text rather than the whole TextField node as sometimes the node is only partially covered, and during animations where the timing of when to hide the textfield was tricky to get right. I ended up subclassing CCTextField and overriding 'update' to honour a new property of 'just hide the UITextField' after calling super.

twomedia:
Noticed this issue still exists when building for OSX, just copied @RobBoluga solution over to the CCPlatformTextFieldMac.m (#1373)

PR #1180, opened 2015-01-26
Fix for Issue #1122

RobBoluga:
See also this post int the Forum:
http://forum.cocos2d-swift.org/t/cctextfield-textfield-property-is-alway
s-nil/15495


iMobCoding:
Can somebody merge at least lines that return _textField:

```
- (id)nativeTextField
{
    return _textField;
}
```

It's pretty urgent for me to access native UITextField and I don't want to mess with the original cocos lib...

Thanks

PR #1181, opened 2015-01-26
More efficient recursion when converting to points

RobBoluga:
convertContentSizeToPoints and convertContentSizeFromPoints now avoid
unnecessary recursion to improve efficiency

### DIFF
--- a/cocos2d/CCNode.m
+++ b/cocos2d/CCNode.m
@@ -426,26 +426,20 @@ TransformPointAsVector(CGPoint p, CGAffineTransform t)
     }
     else if (widthUnit == CCSizeUnitNormalized)
     {
-        if (!gotParentSize) {
-            gotParentSize = YES;
-            parentsContentSizeInPoints = _parent.contentSizeInPoints;
-        }
+        gotParentSize = YES;
+        parentsContentSizeInPoints = _parent.contentSizeInPoints;
         size.width = contentSize.width * parentsContentSizeInPoints.width;
     }
     else if (widthUnit == CCSizeUnitInsetPoints)
     {
-        if (!gotParentSize) {
-            gotParentSize = YES;
-            parentsContentSizeInPoints = _parent.contentSizeInPoints;
-        }
+        gotParentSize = YES;
+        parentsContentSizeInPoints = _parent.contentSizeInPoints;
         size.width = parentsContentSizeInPoints.width - contentSize.width;
     }
     else if (widthUnit == CCSizeUnitInsetUIPoints)
     {
-        if (!gotParentSize) {
-            gotParentSize = YES;
-            parentsContentSizeInPoints = _parent.contentSizeInPoints;
-        }
+        gotParentSize = YES;
+        parentsContentSizeInPoints = _parent.contentSizeInPoints;
         size.width = parentsContentSizeInPoints.width - contentSize.width * director.UIScaleFactor;
     }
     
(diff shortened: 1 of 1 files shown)


andykorth:
I don't really think this is an improvement because:

On lines 429, 437, 445, `gotParentSize` is checked, but it is always `NO` at each one of those points. So those if blocks always execute. Then, the only savings occurs that a local variable is potentially used in the height calculations instead of looking at `_parent.contentSizeInPoints`, but that only occurs once. 

`parentsContentSizeInPoints` could just be filled at instantiation on line 416, but that just converts two (or fewer) lookups to one, so I don't really think that would be much of an improvement.

What sort of percentages of improvements did you see in the profiler with these changes?

RobBoluga:
You are right about the checks on lines 429, 437 and 445, and similar checks in `convertContentSizeFromPoints`. These can be removed.
However, the change does avoid some exponential recursion as `parentsContentSizeInPoints` will again call `convertContentSizeToPoints` for any parent nodes. This effect can be seen if you nest several nodes and set their content size as relative (normalised).
We setup a test project with MainScene including 6 CCNodes nested within each other. The root node has a content size of 100% and each child at 50% of it's parent. With the current version of Cocos2d, calling `contentSizeInPoints` on the lowest node in the hierarchy results in `convertContentSizeToPoints` being called 485 times. With our change it is called 68 times.

andykorth:
Thanks for the quick reply, I don't think I realized which methods exactly were recursive. Thanks for making those fixes!

RobBoluga:
Not too scientific, but to further illustrate the performance improvement, we added 6 calls to the lowest node's `contentSizeInPixels' to the update method in MainScene.
We then ran it on a 5S without our changes and got an FPS of 22 to 25. Running the same project with our changes maintained 60 FPS on the same device.
We've also seen big performance gains like this in 2 of our major projects. We believe anyone using nested nodes with relative positioning will see improvements.

RobBoluga:
I've committed the removal of the redundant `if()` statements. Thanks.

RobBoluga:
We've found a further improvement by implementing the same cacheing principal to `convertPositionToPoints` and `convertPositionFromPoints` so I've committed these changes too.

andykorth:
I'm taking a look at merging this today. Will write some unit tests

andykorth:
Here are some timed unit tests before the change:

```
Test Case '-[CCNodeTests testPerformanceInstantiatingNormalizedNodes]' passed (0.016 seconds).
Test Case '-[CCNodeTests testPerformanceManipulatingNormalizedNodes]' passed (0.080 seconds).
Test Case '-[CCNodeTests testPerformanceCalculateNodeToParentTransform]' passed (0.246 seconds).
Test Case '-[CCSchedulerTests testLotsOfRepeatingTimers]' passed (8.851 seconds).
Test Case '-[CCSchedulerTests testLotsOfTargetRemovals]' passed (4.806 seconds).
Test Case '-[CCSchedulerTests testLotsOfTimers]' passed (2.701 seconds).
```

and after:

```
Test Case '-[CCNodeTests testPerformanceInstantiatingNormalizedNodes]' passed (0.016 seconds).
Test Case '-[CCNodeTests testPerformanceManipulatingNormalizedNodes]' passed (0.080 seconds).
Test Case '-[CCNodeTests testPerformanceCalculateNodeToParentTransform]' passed (0.192 seconds).
Test Case '-[CCSchedulerTests testLotsOfRepeatingTimers]' passed (7.721 seconds).
Test Case '-[CCSchedulerTests testLotsOfTargetRemovals]' passed (3.774 seconds).
Test Case '-[CCSchedulerTests testLotsOfTimers]' passed (2.669 seconds).
```

Out of all of the tests, the nodeToParentTransform calculations are probably the most representative of the gains a normal app would see with a lot of nodes that are in a fairly deep hierarchy. nodeToParentTransform uses these conversions and tends to get called every time something is drawn.

Also some very nice improvements on repeating timers and action removal tests... basically anything that touches the transform.

andykorth:
Thanks for the pull request! Looks good to merge here!

PR #1182, opened 2015-01-26
Ensure animation always reset

cocojoe:
Remove caching


cocojoe:
Have tested in both Cocos2D and SB tests.  Also had it community tested.  Although perhaps not 'critical' it can be subtle annoyance depending on your animation use case.

Issue #1184, opened 2015-01-26
Name collision in CCParticleSystemBase


andykorth:
It uses a `BOOL active`, which is used by CCNode. As a result, particle systems tend not to start.

Fix incoming.

slembcke:
Duplicate of #1178

PR #1185, opened 2015-01-26
Fix name collision on particles

andykorth:
Fix issue #1184


slembcke:
This was caused by renaming `CCNode.isRunningInActiveScene` to just `CCNode.active`.

The alternative is to come up with a different name for `CCNode.active` so that the particle property can remain backwards compatible.

slembcke:
I'm a little worried about this causing confusing backwards compatibility bugs. Code using the old CCParticleSystem.active property would compile without warning, but its behavior would change. It's probably a rare property to use, and after some discussion, changing either it's definition or the name of the CCNode.active property isn't really a better solution. Will merge this as is now.

PR #1190, opened 2015-01-28
Renamed CCFileUtilsV2 and related classes to CCFileLocator.

NickyWeber:
Incorporated some comments in https://github.com/cocos2d/cocos2d-swift/pull/1173.
- Renamed new file utils class to CCFileLocator
- Renamed `imageNamed` to `fileNamedWithResolutionSearch`
- Added test cases again to UnitTests target, got lost


andykorth:
Good catch noticing the tests. These renames look good to me.

PR #1191, opened 2015-01-28
Mac CCResponder refactoring

andykorth:
Refactoring the Mac input code with these goals:
- Reduce number of similar mouse event types sent to make it more uniform.
- Clean up `[CCDirector bind:]` calls- there were too many before due tot he copy-pasted code
- Reduce the amount of duplicated code
- Add support for mouseMoved events. 

The API for handling mouse events has been reduced from 9 methods to 3. Before, we had `mouseUp` (left mouse button), `rightMouseUp`, and `otherMouseUp`, etc. Now, `mouseUp` is passed a CCMouseButton. I'd like feedback on that change.

The CCMouseButton enum was moved between headers. I considered removing it in favor of the already existing NSEventType enum. However, I decided against changing the existing code for compatibility and because those events are sort of complicated, they encapsulate both mouse button and state: `NSLeftMouseDown`, `NSOtherMouseUp`, etc. Thoughts on that would be appreciated also.

In this pull request, the commits are well organized, so you might want to review it by going through each commit one at a time. The purposes of each change should be quite clear that way.


cocos2d-ui/CCSlider.m
```diff
@@ -133,13 +133,13 @@ - (void) mouseUpOutside:(NSEvent*)event
     [self inputUpOutside];
 }
 
-- (void) mouseDragged:(NSEvent*)event
+- (void) mouseDragged:(NSEvent*)event button:(CCMouseButton) button
 {
     CGPoint dragPos = [event locationInNode:self];
     
     [self inputDraggedWithPos:dragPos];
```
slembcke, line 144: This needs an if(button == left) check.

slembcke:
Is there anything better we can do than "other" mouse button? It seems really odd after using it.

PR #1192, opened 2015-01-28
More efficient recursive positioning of cc node

andykorth:
See: https://github.com/cocos2d/cocos2d-swift/pull/1181

I've added some unit tests for that change.

### DIFF
--- a/UnitTests/CCNodeTests.m
+++ b/UnitTests/CCNodeTests.m
@@ -500,6 +500,68 @@
 	
 }
 
+-(CCNode*) makeChainOfNodes
+{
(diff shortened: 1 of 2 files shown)


slembcke:
Everything looks correct. Parent relative position types have shown up in the profiler for me on a few occasions as well. Seems like this makes a significant difference.

The internal implementation of the position types will be changing in v4. We are making it so that there are explicit layout events instead of recalculating the values repeatedly. This should provide a nice speedup for 3.4 in the meantime though.

cocojoe:
Great was just about to create an issue for this.  Thx

PR #1202, opened 2015-01-29
CCBReader updated for new blendmode written by Spritebuilder.

NickyWeber:
Updated binary reader for new blend mode written by spritebuilder.

In conjunction with SB pull request: https://github.com/spritebuilder/SpriteBuilder/pull/1324

### DIFF
--- a/cocos2d-ui/CCBReader/CCBReader.m
+++ b/cocos2d-ui/CCBReader/CCBReader.m
@@ -736,14 +736,10 @@ static inline float readFloat(CCBReader *self)
 
         if (setProp)
         {
-            [(id <CCBlendProtocol>) node setBlendMode:[CCBlendMode blendModeWithOptions:@{
-                    CCBlendFuncSrcColor : @(readIntWithSign(self, NO)),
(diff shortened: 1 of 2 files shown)


slembcke:
√

andykorth:
Maybe I'm super confused, but it looks like those blend mode changes are in a flipX and flipY block. They should be down a block?

Also, I'll be that guy: looks like the indentation inside each #if DEBUG_READER_PROPERTIES preprocessor block is wrong. _shrugs_. Maybe we can get that later all at once.

slembcke:
Hmm. Yeah. It's a little more clear when the diff context is expanded.

NickyWeber:
@slembcke Sorry that was my mistake, I had it set up correctly in my SB project but copy and paste was a missed target. Pushed a correction.

How should the the `#ifdefs` look like? I think that I indented the code it encapsulate with the same level as the surrounding code.

rkachowski:
Is it possible to merge this? 

Currently the writer version is out of sync with the reader, and spritebuilder will generate files that it can't read :/

andykorth:
Let's definitely not delay it about indentation.

PR #1204, opened 2015-01-30
Bugfix/headless

slembcke:
Pulling in @rkachowski's Android fixes for the latest v4 changes. I'll merge this in the next couple days unless there is any feedback since the changes are pretty Android specific.

Is there still a strong reason to have a separate cocos2d-tests-android project? I feel like the reason it was separate no longer exists, but I haven't had time to look into it. It would certainly reduce the effort to test on Android more regularly. I think I started testing on Mac 3-4x more often once the iOS/Mac test projects were merged. Any volunteers that want to tackle this?

There is still one major issue, touch input doesn't seem to work at all on Android. `[CCGLView onTouchEvent:] isn't being called at all, so there seems to be something wrong with the view's setup. There is only about a dozen lines of code that interact with the view at setup time, and it hasn't really changed from 3.4. I gave up on trying to figure it out tonight.


rkachowski:
@slembcke wrt regularly testing the android suite, our third party QA team has instructions to run this project on every test sweep - resulting in issues like  ( https://github.com/spritebuilder/SpriteBuilder/issues/1321). There's no real reason to have it separate afaik and we could merge it with whatever other functional integration tests we have for other platforms.

rkachowski:
I fixed the touch input here and created a pull request (i can't push to this branch) with all the changes combined  - https://github.com/cocos2d/cocos2d-swift/pull/1205 - It looks like the BRIDGE_CLASS macro got ran over in the v4 merge. 

Touches work now, but the menu is pretty funky. I think this is a project / application matter in Headless.xcodeproj rather than a framework issue.

slembcke:
Replaced by #1205.

PR #1205, opened 2015-01-30
Compatibility fixes for Android with v4

rkachowski:
A few problems prevent android from being run on the newest version of cocos, and there have been some issues stacking up on the spritebuilder repo. This PR is designed to address that.

All the code here should only significantly affect the Android platform, and is designed to bring everything into line with the new director changes.

_NOTE_ - base64 support is still not available and will need some love in the very near future.


slembcke:
I made issues #1206 and #1207 for the remaining issues.

PR #1210, opened 2015-02-02
Fix bad line splitting on updateLabel.

sarah-j-smith:
See bug report and demo project: https://github.com/sarah-j-smith/CCLabelBMFontBug


slembcke:
Good catch. We should add a couple more tests. Double check that a word longer than the width doesn't explode.

andykorth:
Firstly, a huge thank you for the replication project! Yay!

So overall this change is an improvement, although there are some potential odd cases:
Trying some long words with the old code:
Before

And with the new code:
Before

We have proper word wrapping, but the text can extend past the bounds. (I've hardcoded the width in these two examples to be the same).  So while I think this is probably an improvement, I'm not sure if we should merge it as-is, since some people might rely on the old behavior of hard limits on the width of lines. Because of how `lastWord` is stored, I guess long lines can cause extra newlines. 

Maybe BMFont needs a wordwrap toggle that switches between these behaviors.

(I think ideally, it would word wrap, unless the word is longer than the line width, then it would character wrap)

sarah-j-smith:
Glad the replication project helped.  It was kind of fun to make.  I liked "disco" and "intent".  Grin.

The comments from the original author of the code were:

```
//Character is out of bounds
//Do not put lastWord on current line. Add "\n" to current line to start a new line
//Append to lastWord
```

But the code wasn't doing that.  My change actually does make a change to `multilineString` which is what the comment says was the intent.  And it fixed the issues I was seeing in my game.  :+1: 

My use case is that I have a "cartoon speech bubble" class that displays in game hint messages which I can control.  Those were getting split in weird ways, and it seemed to be a regression, as it used to work.

But I agree we should look to see if this can be made better.  Is true that some options for wrapping would be good.

Also agree that the _word-wrap unless longer than line width then character wrap_ sounds like a great default 90% case.

I guess for a situation where you're displaying strings from somewhere outside of the game: a player entered string, or something from the web, then the long word issue could bite.  Also some translation languages might cause an issue, especially for very small labels.

My change was basically to make what looked like the original intent of the Cocos2D developer actually work - in other words this looked like a bug and a regression to me.  

Maybe we could make a new issue to improve this still further?

andykorth:
Haha, yeah, I liked the "winter of our disco"- figured that might be on purpose :D

You're right about it being a regression, it looks like it was introduced here:
https://github.com/cocos2d/cocos2d-swift/pull/1050

And here's the discussion: https://github.com/cocos2d/cocos2d-swift/issues/1018 with some more in the forum: http://forum.cocos2d-swift.org/t/cclabelbmfont-bug-with-newlines/15408

andykorth:
After a bit more looking, I think we should merge this and make a new issue for character wrapping long words.

sarah-j-smith:
OK, I have done a lot of hacking on this, with the following results:
- long words now correctly split
- it will split on word breaks first
- children of the label don't get hidden
- you can now address sprites
- width wrap doesn't cause \n to be inserted
- the code is IMHO now maintainable
- there are tests for the width wrap 

Screenshot of the new tests which exercise the `width` setting along with alignment.  The previous tests did not test either of these things.
img_0013

The current doc says 

_Individual characters are sprites, can be modified individually (ie individual character animations)._ 

...but you can't do that at present because the rather opaque `skip` and `justSkipped` logic that uses the `getChildByTag` method means there is no mapping between children and sprites.  This especially so because the old way of handling the wrapping when the `width` property was set was to insert actual line break characters into the string.

Now what happens is the CCSprite private sub-class handles the iteration over sprites, and hard & soft line feeds are handled by meta-data on the sprites (an enum property on the sub-class).

To fulfil the promise of the doc I've added a `characterSpritesForRange:` method (tested and documented) which returns a list of the sprites so you can apply effects to them.  Here's some screenshots of those tests:
img_0018
Setting color on the sprites.
img_0019
Setting effects on the sprites.

andykorth:
This looks great! The bitmap font code really needed a cleanup.  The tests are a big help too. Do we want this merged into v3 or just for v4?

slembcke:
Wow... Awesome. This looks a lot better. There was a lot of cleanup to be done in this class that I was dreading to do. Everything seems to be in order here. : D

PR #1214, opened 2015-02-04
Fixed calculation of tangent/binormal vectors.

slembcke:
This is a relatively major bug in the lighting effect. The code was inverting the node-to-local matrix, effectively passing the reverse of the object's transform to the lighting effect. This would undo the rotation or skewing of an object in the lighting calculations. Only the position of the sprite affected the lighting calculations.

I think this fixes the intent of the original code, but I don't see any tests that use normal maps that are asymmetric. I'm not 100% certain that this is correct without an extra y-flip or 180 degree rotation or something silly like that.

### DIFF
--- a/cocos2d-ui-tests/tests/CCEffectsTest.m
+++ b/cocos2d-ui-tests/tests/CCEffectsTest.m
@@ -327,6 +327,8 @@
         sprite.effect = lightingEffect;
         sprite.scale = 0.5f;
         
+        [sprite runAction:[CCActionRepeatForever actionWithAction:[CCActionRotateBy actionWithDuration:1.0 angle:90.0]]];
+        
         CCNode *root = [[CCNode alloc] init];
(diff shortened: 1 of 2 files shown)


thayerandrews:
Thanks Scott. I will take a look at this today.

thayerandrews:
Looks good. Nice catch! I can't believe I didn't have a test for this yet.

I don't have the time right now to dig further into what's wrong but, FWIW, I was trying to transform the local tangent (1, 0) by the normal vector transformation matrix that fixed-function GL uses for lighting (http://en.wikipedia.org/wiki/Normal_(geometry)#Transforming_normals). I'm not sure why that didn't work but since your solution works and is simpler and faster I say, "Ship it!"

BTW, I will submit a pull request ASAP that includes a non-radially symmetric normal map that we can use for testing.

PR #1218, opened 2015-02-05
Update Test project to work with CCAppController changes

andykorth:
Test changes for https://github.com/cocos2d/cocos2d-swift/pull/1217

This pull request includes the required changes to get the visual tests working with the new CCAppController changes. It includes a few minor changes I'll comment on in https://github.com/cocos2d/cocos2d-swift/pull/1217. 

Tested on Mac/simulator/iOS. Most of the work was making sure search paths and other configuration settings were correct for the somewhat odd setup of the tests (compared to spritebuilder projects)


andykorth:
This looks great. I've updated the test projects in https://github.com/cocos2d/cocos2d-swift/pull/1218 and I made a few minor changes in that pull request. 

[CCBReader configureCCFileUtils]; overrides the search paths that were set up by user code in setupApplication, so I have moved that call into `iosConfig`, where it can be overridden and avoided, if necessary (by not calling `super`). See https://github.com/andykorth/cocos2d-iphone/commit/c0f043d0323db9ffb4e18b7cf7a18a53475f886d#diff-3d6f4834781c4d5319984b04a010010bR107

For whatever reason, the testbase doesn't have a `"Published-iOS"` folder, so it can't find the `configCocos2d.plist` file. Not sure if this ever worked with the testbase. Probably unimportant.

When config files are missing, the default `iosConfig` returns a nil `config` dictionary. Which is of course, silently ignored when you try to add things to it. I feel that should either be an error, or it should give you a empty dictionary if there isn't a plist. I've implemented the latter behavior. See https://github.com/andykorth/cocos2d-iphone/commit/c0f043d0323db9ffb4e18b7cf7a18a53475f886d#diff-3d6f4834781c4d5319984b04a010010bR100

I definitely agree about encapsulating the director stack calls. I had a slightly different idea, see if you like it.
The user writes this:  [Line 73, user's AppController](https://github.com/andykorth/cocos2d-iphone/blob/32820215afb6e423c3ea1b5ed60d3415cc6e4c62/cocos2d-ui-tests/ios/AppController.m#L73)

To override the default behavior here: [Line 78, CCAppController.m](https://github.com/andykorth/cocos2d-iphone/blob/c0f043d0323db9ffb4e18b7cf7a18a53475f886d/cocos2d/Platforms/CCAppController.m#L78)

If they don't override `createFirstScene`, then it works as in your initial commit, where they only need a string.

rkachowski:
You're right about the config, it should return a valid dictionary if the file doesn't exist. 

Incidentally there is no file in the testbed as the testbed isn't a SpriteBuilder project - sometime after v4 release we should probably recreate these functional tests to use 1) the new project format and 2) some kind of cool automation framework (but this is kind of irrelevant here).

> See andykorth@c0f043d#diff-3d6f4834781c4d5319984b04a010010bR107

Yeah that's a good idea, I wasn't aware of `configureCCFileUtils` running over everything. The same changes will need to be made for the other platforms though.

> The user writes this: Line 73, user's AppController

Yeah that's cool, I exposed and documented these methods in the header to get this effect. In practice I'm not sure how often it'll be overridden by a user though.

andykorth:
I'll change the other platforms, then this will be ready to merge!

andykorth:
@slembcke  Ah, I replicated and fixed the issue with the Testbase on Mac. The 582049b commit was a minor change to make android and iOS consistent, but it introduced an extra `[CCBReader configureCCFileUtils]` by accident. That resets user-set paths, and was causing the Mac issue. Android tests might have the same problem.

Mac is fixed with the above commit.

slembcke:
Looked into the iOS 7 issue more. The problem is a difference in how layout events happen. On iOS 8, when you call `[UIWindow makeKeyAndVisible]` it seems to defer setting up the layout until it's drawn to the screen. On iOS 7, it forces the layout immediately. This is bad since the the layout reshapes the projection, and reshaping the projection is still the current place where the initial scene is initialized. Since the rest of Cocos hasn't been configured yet, it crashes.

I'll see if we can figure out a workaround for now, though I'm hoping to get rid of CCAppDelegate completely before we release v4. That should help simplify things a lot.

slembcke:
Hmm. Can't figure out the Android issue. We can't get the SB templates to run when we pull the latest Cocos2D changes. I put logging statements in all of the activity methods and none of them run. I double checked that com.apportable.GLActivity was set in the manifest file, and that the template's activity class was bridged to it. That's the limit of my knowledge on how the Android system starts up.

rkachowski:
> We can't get the SB templates to run when we pull the latest Cocos2D changes.

You are of course referring to this pull request, and not latest cocos2d develop? As the templates only exist in https://github.com/spritebuilder/SpriteBuilder/pull/1328 which needs this PR here to run

I'm going to copy this list from the previous PR
1. Checkout https://github.com/spritebuilder/SpriteBuilder/pull/1328 on SpriteBuilder
2. Go to the cocos submodule and checkout this pull request
3. Remove `Generated` folder from project root if it exists
4. cd to `scripts` and run `./build_distribution.py --version 1.4`
5. Once the template is generated, run SpriteBuilder and create a new project
6. Publish, open in XCode and test different platforms. (requires the most recent SBAndroid plugin)

rkachowski:
I've managed to reproduce the issue on iOS 7, I hope to have some changes to address this soon which @andykorth will need to pull into this pull request as I no longer have control over it.

rkachowski:
The iOS 7 fix appears to be to load the `.ccb` directly in `-(void) directorDidReshapeProjection:` within `CCAppDelegate`. It appears that the `CCBReader` has a dependency on `UIWindow#makeKeyAndVisible` returning successfully. Loading and instantiating a ccb file before that results in corrupted layout issues.

You can also remove this (hacky) callback method completely from CCAppDelegate and instead explicitly call `[director presentScene:[self startScene]` at the end of setupIOS. 

This is more consistent and works towards removing the CCAppDelegate file but unfortunately adds a frame of flicker between the Default.png and the first scene on iOS 8.  Regardless of whatever solution you choose there is flicker on iOS 7 (it exists on SpriteBuilder develop + cocos2d develop).

I don't have write access over this pull request so I kindly ask one of you guys, @andykorth @slembcke to affect these changes

andykorth:
That fix sounds good. I'm downloading the iOS 7 simulator, and I'll give it a try and add it to this pull request. Thanks for looking into it. It was very strange for iOS to create and set up the first scene inside `directorDidReshapeProjection` anyway, so I consider this an improvement regardless. 

Any idea with the android issue? We definitely had the templates set up properly yesterday, as the other platforms were working. 

I think we can figure out the flickering issues later, once the refactoring is done. The old "flickr" hacks were very strange.

slembcke:
@andykorth  Err, wait. Did we not commit the iOS 7 fix yesterday? We did something similar, basically moving the `[UIWindow makeKeyAndVisible]` call.

andykorth:
Oh hm, where did that go to? It's not in this branch.

rkachowski:
> Any idea with the android issue? We definitely had the templates set up properly yesterday, as the other platforms were working.

I don't know what to tell you, it works fine with me. Could you tarball me a project that you guys are experiencing issues with on Android and I'll have a look?

andykorth:
I ended up fixing iOS 7 support as Donald suggested. iOS 7 and 8 are working now. For what it's worth, I'm not seeing any flickering on either simulator or my iOS 8 device with the code in 60df1c9aa20bcd58b3862faeff2f133c5148638e.

slembcke:
Ok. So I've been able to get the templates to run using https://github.com/spritebuilder/SpriteBuilder/pull/1328. I'm going to see if I can figure out what the difference is that's preventing it from working with the tests project. Then I'll merge.

slembcke:
Wasn't able to figure out the Android testbed issue. Since the templates work, we know it's fixable. I don't want to delay merging this any more since it's holding a lot of other work up.

PR #1224, opened 2015-02-11
CCEffectsTest - Add a test of lights and render textures

thayerandrews:
This test demonstrates a lighting limitation that a user encountered and reported in the forums. 

When drawing to a render texture and using manual calls to begin, visit, end, the visited node graph must be part of the scene. If this is not the case then the lighting effect cannot find the scene node and the active light collection.


slembcke:
Ah. That explains why it works in Galactic Guardian then. I render the scene itself into the render texture. We had discussed an idea for the renderer object to have a dictionary of temporary rendering values. Would this be helpful in solving the problem? It would be trivial to implement. We'll still need to add something to the public API though.

This should probably go into the v3.4 branch instead if that's where the bug fix will go. I can cherry pick the the commits if you want.

thayerandrews:
In this user's case, I think the temporary rendering values dictionary wouldn't help since both the lit node and the light are in the manually visited subtree.  Because lights add themselves to the light collection in their onEnter method, the light still wouldn't be in the light collection when the user calls visit. If the light were already in the scene though and it was just the lit node that was being manually visited then the temporaries dictionary would fix the problem.

I do like the idea of the temporaries dictionary as a means to access current rendering state that is otherwise difficult to access. What other use cases do you have in mind for it?

And I think you're right that this should go into v3.4. I'll close this pull request and do the cherry-pick.

slembcke:
Right... Hmm. Not really an easy solution here then. I guess the workaround is to add the node to the scene using a hidden parent. Then it will get the lifecycle messages, but ever be visited.

I never implemented the renderer dictionary because I never had a real use case for it. The implementation would be trivial, but I didn't want to add it speculatively.

thayerandrews:
Yeah, I think it's just something I need to document as a known limitation
of effects and render textures.

There are some other oddities like this one when users combine effects and
the manual visitation subgraphs. For instance, in reflection and
refraction, finding the transform from the affected sprite to the
environment sprite is not well defined if the user manually visits an
intermediate node in the hierarchy (i.e. an ancestor of the affected node
that is not an ancestor of the environment). I've never come up with a good
solution to that one so I just need to document it as a will-not-work
situation.

thayerandrews:
Okay, I've submitted a new pull-request that brings this over to v3.4.

slembcke:
Yeah... the explicit visit calls have always been sort of a problem. There is a _lot_ of undocumented or undefined legacy behavior. I tried to maintain as much of it as possible when I implemented CCRenderer. I've never had anything other than vague ideas for improvements though.

PR #1226, opened 2015-02-11
CCEffectsTest - Add a test of lights and render textures

thayerandrews:
This draws a lit sprite without a render texture, with an auto-drawing render texture, with a
render texture and manual visitation of a sprite that is not a child of the RT, and with a render
texture and manual visitation of a sprite that is a child of the RT. Cases 1, 2, and 4 should draw
correctly while in case 3 the sprite should be all black instead of being lit.


thayerandrews:
Thanks!

PR #1228, opened 2015-02-12
v3.4 class reference updates

LearnCocos2D:
Only code change, added missing initializer:
https://github.com/LearnCocos2D/cocos2d-swift/commit/4ba378011e8cddf234e3f9506ab95e4516990d7c


LearnCocos2D:
PS: you may want to merge this req into v3.4 instead:
https://github.com/cocos2d/cocos2d-swift/pull/1229

PR #1230, opened 2015-02-12
Expose various CCNS methods across platforms.

rkachowski:
This pull request is designed to allow cocos2d to compile on mac, as the `CCNSStringFromCGPoint` is being used across platforms as of https://github.com/cocos2d/cocos2d-swift/commit/78ed94651d7504c7d5f3dc1165d9854481f8dbf8, despite the unavailability of the method on Mac. This prevents applications which rely on cocos2d (i.e. SpriteBuilder) from compiling.

Also changed here is the correct use of the header guards and removal of a weird `#if 1` statement.

### DIFF
--- a/cocos2d/Platforms/CCNS.h
+++ b/cocos2d/Platforms/CCNS.h
@@ -26,11 +26,9 @@
 
 #import "ccTypes.h"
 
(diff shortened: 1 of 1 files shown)


slembcke:
I could swear I tested the Mac build after merging the BMFont pull request yesterday. I guess I didn't. Sorry!

PR #1235, opened 2015-02-16
CCEffectLighting - Fix alpha blending issues

thayerandrews:
- Ensure that lighting only affects the RGB channels and does not change alpha.
- When adding the specular term to the final result color, multiply it by the input color's
  alpha value so everything stays in the premultiplied alpha space.


thayerandrews:
Before this fix, lighting around alpha blended edges might look something like this:
lighting alpha broken
When instead it should look like this:
lighting alpha working
The specular contribution of the lighting equation was not correctly accounting for premultiplied alpha space. It seems I will never write an initial effect implementation without making this error and having to go back and fix it. Blerg.

PR #1238, opened 2015-02-18
CCEffectLighting - Add overall lighting intensity control

thayerandrews:
The property ranges from [0..1]. If the value is 0, lighting has no effect on
the affected sprite. If the value is 1, lighting contributes fully to the affected
sprite. Values in between allow addition of some lighting to a sprite.

### DIFF
--- a/cocos2d-ui-tests/tests/CCEffectsTest.m
+++ b/cocos2d-ui-tests/tests/CCEffectsTest.m
@@ -366,25 +366,26 @@
     setupBlock(ccp(0.75f, 0.5f), CCLightDirectional, 0.5f, @"Directional Light\nPosition does not matter, orientation does.");
 }
 
--(void)setupLightingEffectIntensityTest
+-(void)setupLightingEffectContributionTest
 {
-    self.subTitle = @"Lighting Intensity Test";
+    self.subTitle = @"Lighting Contribution Test";
     
     [self.contentNode.scene.lights flushGroupNames];
   
     NSString *normalMapImage = @"Images/powered_normals.png";
     NSString *diffuseImage = @"Images/powered.png";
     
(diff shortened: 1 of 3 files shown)


slembcke:
Looks fine, though should it also affect the ambient factor?

I am starting to wonder if we are starting to have too many variables that all control basically the same thing. There are now 5 variables that are all multiplied together to get the overall brightness right? (light color, light intensity, sprite color, sprite texture, effect intensity)

slembcke:
Talked with Viktor a bit about this. I should have run it instead of skimming the code. -_-

I think we should definitely call it something other than "intensity" since it's very different from the CCLightNode.intensity property.

I'm still not 100% sold on the idea, but I wouldn't veto it. It just seems like a very specific use case.

thayerandrews:
You're right about the ambient contribution. I'll fix that. And the name intensity is also too similar to light node intensity. Would "contribution" be better?

Regarding the specificity of the use case, that's a fair point. It makes me think back to an idea Oleg and I originally discussed and then didn't pursue which was to make effect stacks more flexible with some sort of "combiner" effects. They could take the results of multiple input effects (not just the previous one) and add them, multiply them, lerp them, etc. That would be nice here since this intensity factor would no longer need to be a property of CCEffectLighting and could instead be a property of a separate effect that controls how lighting is combined with the sprite's color. 

There's a lot more to it than this of course (which is why we left it alone at the time) but it's interesting food for thought.

thayerandrews:
Ooops. I need to pay more attention to my own code (and not post comments when I'm tired). It turns out the ambient contribution is already handled in the original change. Within the GLSL, the global ambient term feeds into diffuseSum and then diffuseSum is influenced by u_overallIntensity so everything just works. I'm adding a test case that exercises this though.

thayerandrews:
I shouldn't have submitted this PR from my v3.4 branch. Will resubmit from a topic branch in the future.

PR #1240, opened 2015-02-18
Fix cclabelbmfont

sarah-j-smith:
Bring fix-cclabelbmfont fixes into v3.4. 
- Fixes font wrapping and width fitting/line splitting
- Fixes bad content size calculations
- Fixes missing characters 
- Allows addressing of font sprites for effects, colour

Add tests for FooBar.bmfont hierarchy font files, with scaled fonts for phone, phonehd etc.  For this a test font was created and added to the test resources.

**Note that the GPL'ed font has an exclusion so it is OK to use**


sarah-j-smith:
Closing this PR, to be replaced soon by a new one incorporating improvements from this forum discussion: http://forum.cocos2d-spritebuilder.org/t/found-two-new-bugs-in-new-cocos2d-3-4-and-spritebuilder/16712/20

andykorth:
It looks like the issue reported in the forum was fixed in https://github.com/cocos2d/cocos2d-spritebuilder/pull/1257

But what about the other changes in this pull request? It looks like there's a lot of good cleanup in here! Should this be re-opened and merged into develop?

sarah-j-smith:
Hi Andy, I believe all the final fixes for CCLabelBMFont went into develop, via various PR's.  The last one was "Fix for narrow labels" as reported by the poster in the forum.

But yeah, oops - I didn't create a v3.4 PR as I said I would above.  I wasn't sure whether v3.4 was the right place to PR against. 

I cherry-picked those changes into my local fork v3.4 so that the issues reported in the forum (which had a test case build against v3.4) could be tested by the forum poster.  There was a couple of minor merge conflicts which I fixed.  https://github.com/sarah-j-smith/cocos2d-swift/commits/v3.4 

_But_ I don't think those changes got into v3.4 upstream.  Not sure - check with @cocojoe who I think had a go at merging/cherry-picking - if you want me to make a PR from that let me know and I'll click the buttons.  :smile:

PR #1244, opened 2015-02-19
Rename CCB classes and file extensions to CCSB / .sb

andykorth:
The .ccb and .ccbi file extensions have changed to .sb and .sbi respectively. This clarifies that SpriteBuilder is the source of these files.

In addition, a few CCB\* classes have been renamed to CCSB*, which keeps the CC prefix uniform throughout the project and includes SB to note that it deals with SpriteBuilder files.

Compatibility aliases are provided in the deprecated header to ease the transition.

This pull request will have accompanying SpriteBuilder changes, so the two pull requests should be merged at the same time.


andykorth:
The corresponding SpriteBuilder changes can be found at: https://github.com/spritebuilder/SpriteBuilder/pull/1368

andykorth:
After this gets merged, I will bump the pull request for SB here: https://github.com/spritebuilder/SpriteBuilder/pull/1368 Then after those tests pass, that pull request can be merged.

PR #1251, opened 2015-02-22
FIX: changed CCTableView's dataSource property from 'strong' to 'weak'

zbarryte:
This prevents retain-cycle that normally occurs when using a SpriteBuilder Scene (CCNode that's 0th child of CCScene generated via loading from CCBReader) as the dataSource of a CCTableView


slembcke:
Hmm. Interesting comparison. UITableView's data source property is weak also. I worry a little bit that since this changes the behavior of the property it will cause compatibility problems for some people. I'll consider merging it into the v4 branch though.

Let me get back to you.

PR #1257, opened 2015-03-01
Fix for narrow labels.

sarah-j-smith:
A bug in the word-wrapping logic manifests on very narrow labels:

http://forum.cocos2d-spritebuilder.org/t/found-two-new-bugs-in-new-cocos2d-3-4-and-spritebuilder/16712/20


slembcke:
I didn't add a test specifically for this issue, but I made a note for it elsewhere. Seems simple enough so I'm just going to merge it as is.

PR #1263, opened 2015-03-10
Animation bug fixes and compile fix

andykorth:
Fix compilation error in CCTransition. Fix release of a property (although maybe it should be auto-synthesized..?)


andykorth:
I also added some tests for actually loading a ccb file (the first in this testbed). We also run the animations from the ccb file. 

Note that the last test I added does demonstrate that the change here: https://github.com/andykorth/cocos2d-iphone/commit/79199b55cfb782fcedc0dafaaaf16a5d4bdf12ef#diff-b132d6e5c0d197fdffbca133b6235978R199 
is necessary. With just `[_animationManager release];` the last test will crash.

PR #1266, opened 2015-03-13
Split several classes out of CCEffect_Private.h into their own files

thayerandrews:
There is no functional change here. It's just a move of various classes since CCEffect_Private.h and CCEffect.m were getting a little bulky.


slembcke:
Apparently this got missed. :-\ Whoops.

thayerandrews:
Thanks Scott! No worries about the delay. It wasn't holding me up.

PR #1269, opened 2015-03-14
Fix Metal rendering performance halts.

### LINKED ISSUE #1034: Metal Renderer: experiencing halts
I tried the new metal renderer.
I am excited about the performance as well as with the fast adoption of this new technology into cocos2d.
Thank you.

I couldn't find much about other people trying it neither in the forum nor here,
so I thought you might be interested in some feedback.

There were some minor issues to get it running with my current project.
e.g. I had to figure out how to compile the metal shaders and
had to include the last changes of CCResponder in the MetalView.m
Finally I got it running, and it is really fast.
But then I experienced some halts from time to time,
when about .5 seconds long, which I hadn't had so far with OpenGL.

Further investigation showed that most of the time the renderer hung in
        drawable = [self.metalLayer nextDrawable];
waiting for a signal.
But there were also

[renderEncoder drawIndexedPrimitives:MetalDrawModes[_mode] indexCount:_count indexType:MTLIndexTypeUInt16 indexBuffer:indexBuffer indexBufferOffset:2__firstIndex];indexType:MTLIndexTypeUInt16 indexBuffer:indexBuffer indexBufferOffset:2__firstIndex];

and
    _currentCommandBuffer = [_commandQueue commandBuffer];

on top of the call stack, when I hit the stop button.
There are a lot of nodes to be drawn in my project and also some background action,
which didn't bother the OpenGL-renderer so far.

Maybe this is a bug in the metal framework itself (btw, I haven't tried it with 8.1 yet)
Or something related just to my project (e.g. I use NSRunloopCommonModes instead of NSDefaultRunLoopMode (to use UIScrollview and also use UIKitDynamics (Yes, I know that sounds strange)).
Maybe there is a possibility to fix this, but I have no idea how (maybe it has something to do with memory allocation or buffers which are too small, just a blind guess)

davidtsai:
Fixes #1034

The main thing that fixes this is presenting the drawable through the command buffer instead of calling present on the drawable itself.  New to Metal, but it seems like the CPU and GPU get completely de-synced when the drawable is presented on its own.

The other stuff is just in case to make sure the drawable are released in a timely manner.

With this fix, you can see xcode properly report the FPS whereas before it would show up blank in code during debug.

### DIFF
--- a/cocos2d/Platforms/iOS/CCMetalView.m
+++ b/cocos2d/Platforms/iOS/CCMetalView.m
@@ -64,7 +64,6 @@
 - (void) dealloc
 {
 	CCLOGINFO(@"cocos2d: deallocing %@", self);
-
 }
 
 -(CAMetalLayer *)metalLayer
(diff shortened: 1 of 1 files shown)


slembcke:
Huh. That does appear to fix the issue.

I used to use `presentDrawable:` and I could swear that I changed to `present` when trying to fix this bug... Either way, thanks! This bother has been bugging me for a while.

heiko-henrich:
Thank you,
finally it's working now!

PR #1277, opened 2015-03-25
Director, app delegate and content scale changes for v4.

slembcke:
First of all, I apologize for the size of this pull request. It's all of the recent changes for the app delegate, director, content scale handling. They were all interdependent enough that it was difficult to do any one of the features separately and it's also been held back by getting a working branch of SpriteBuilder working with it.

This needs to be merged at the same time as https://github.com/spritebuilder/ObjectAL-for-Cocos2D/pull/6, and the SpriteBuilder pull request.


slembcke:
On the positive side. Once this is merged, all but one of the major v4 changes will be ready for an alpha version. \o/

Mac/AppDelegate.m
```diff
@@ -1,23 +1,64 @@
...
+}
 
 - (void)applicationDidFinishLaunching:(NSNotification *)aNotification
 {
-    AppController *controller = [AppController sharedController];
-    controller.window = _window;
-    controller.glView = _glView;
-    [controller setupApplication];
-    return;
+#if 0
```
rkachowski, line 17: Is this intentional?

UnitTests/CCFileUtilTests.m
```diff
@@ -1,239 +1,239 @@
...
-
-	CGFloat scale = 0.0;
-	
-	NSString *path1 = [sharedFileUtils fullPathForFilename:@"Images/blocks.png" contentScale:&scale];
-	XCTAssertTrue(scale == 2.0, @"");
-	XCTAssertTrue(path1 != nil, @"");
-}
-
-@end
+//#import <XCTest/XCTest.h>
```
rkachowski, line 8: I would either remove this file or disable this test case within the project file, than commit commented out code to the repo

UnitTests/CCFileLocatorTests.m
```diff
@@ -30,8 +31,11 @@ - (void)setUp
 
     self.fileLocator = [[CCFileLocator alloc] init];
     _fileLocator.searchPaths = @[[self fullPathForFile:@"Resources"]];
-    _fileLocator.deviceContentScale = 4;
     _fileLocator.untaggedContentScale = 4;
+    
+    [CCSetup sharedSetup].contentScale = 1;
```
rkachowski, line 36: This is cool

Mac/AppDelegate.m
andykorth, line 17: Yeah. There are two ways to configure a cocos2d application. In your template code, you can call setupApplication, or you can do the setup manually.

`setupApplication` has a bunch of defaults for setting up a project with the normal spritebuilder settings. If a person wants to override that behavior, they can provide their own code - setting special search paths, doign something funky with the window, other content scale settings... basically anything that involves `[CCSetup createCustomSetup];` instead of `setupApplication`.

Both implementations are provided here so we can test both code paths by flipping the ifdef. Maybe not ideal, but better than having two targets, or something like that.

UnitTests/CCFileUtilTests.m
slembcke, line 8: Oh, I forgot about this. A couple of the tests need to be migrated to the CCFileLocator tests. Most of them will get deleted.

PR #1288, opened 2015-04-03
Remove static analyzer warnings about fontName

richardgroves:
Before the above change a static analyzer build gives warnings due to this method 'Potential memory leak of an object' due to the fontName being a +1 retain count from the CFFontCopyPostScript name and the method name not being one that indicates it is handing over ownership of the object.

screen shot 2015-04-03 at 17 00 23


slembcke:
Good catch. There are so many false positives that I don't use the analyzer as much as I should.

I changed the code a bit though. You can use __bridge_transfer instead, and it's a little simpler. It tells ARC to take ownership of the pointer and that it needs it to be released when it's done using it.

richardgroves:
Cool - figured there would be a proper way to take ownership of the foundation object - but I know very little about that stuff. Thanks for doing it properly!

PR #1319, opened 2015-05-07
Unit tests for CCColor

richardgroves:
Catches the problem that https://github.com/cocos2d/cocos2d-objc/pull/1318 fixes
And provides a starting point for further tests.  Major current functionality not covered is:
- (CCColor*) initWithHue:(float)hue saturation:(float)saturation brightness:(float)brightness alpha:(float)alpha

and none of the internal color space workings are tested.

### DIFF
--- a/UnitTests/CCColorTests.m
+++ b/UnitTests/CCColorTests.m
@@ -1,0 +1,166 @@
+//
+//  CCColorTests.m
+//  cocos2d-tests
+//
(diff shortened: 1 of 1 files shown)


richardgroves:
They also only do platform specific stuff for iOS - so the NSColor for Mac isn't tested, and the CGColor stuff on non-iOS wont run.

PR #1326, opened 2015-06-22
Added new property colorRGBAAsGLkVector4 to CCNode

pkclsoft:
Whilst chasing a memory leak in CCColor, I found that there is a ridiculous number of CCColor objects being created that simply aren't needed.  They are only created to serve the purpose of creating a GLKVector4 structure.

By adding the code in this commit, all of those allocations and garbage collections are no longer needed.

### DIFF
--- a/cocos2d/CCDirector.m
+++ b/cocos2d/CCDirector.m
@@ -251,7 +251,7 @@ static CCDirector *_sharedDirector = nil;
 		[renderer prepareWithProjection:&projection framebuffer:_framebuffer];
 		[CCRenderer bindRenderer:renderer];
 		
-		[renderer enqueueClear:(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT) color:_runningScene.colorRGBA.glkVector4 depth:1.0f stencil:0 globalSortOrder:NSIntegerMin];
+		[renderer enqueueClear:(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT) color:_runningScene.colorRGBAAsGLkVector4 depth:1.0f stencil:0 globalSortOrder:NSIntegerMin];
 		
 		// Render
 		[_runningScene visit:renderer parentTransform:&projection];
--- a/cocos2d/CCNode.h
+++ b/cocos2d/CCNode.h
@@ -974,6 +974,13 @@
 */
 @property (nonatomic,strong) CCColor* colorRGBA;
 
+/** Returns a GLKVector4 structure built using the nodes colorRGBA property.  This can be used to circumvent the creation of throwaway CCColor objects.
+ 
+ @see colorRGBA
+ @see CCColor
+*/
+@property (nonatomic,readonly) GLKVector4 colorRGBAAsGLkVector4;
+
 /** Returns the actual color used by the node. This may be different from the color and colorRGBA properties if the parent
  node has cascadeColorEnabled.
 
--- a/cocos2d/CCNode.m
+++ b/cocos2d/CCNode.m
@@ -1697,6 +1697,10 @@ CGAffineTransformMakeRigid(CGPoint translate, CGFloat radians)
 	[self cascadeOpacityIfNeeded];
 }
 
+-(GLKVector4) colorRGBAAsGLkVector4
+{
+    return GLKVector4Make(_color.r, _color.g, _color.b, _color.a);
+}
 
 - (void) cascadeColorIfNeeded
 {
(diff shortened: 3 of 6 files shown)


pkclsoft:
The second commit adds some code to the CCColor class that tries to ensure that if a CCColor object is modified after creation, the captured CGColor field (if it exists) is released so that a new one can be created the next time it's requested.

pkclsoft:
The third commit is another improvement by reducing the number of times we call CCNode.colorRGBA.

Birkemose:
Looks good. I will see if I can go over it, and merge it morrow

richardgroves:
There is no need to handle the possibility of the color changing (in https://github.com/pkclsoft/cocos2d-objc/commit/9ce83abd862fd72d80bf7af860c3848181b15d34) and needing to release the GCColor object as CCColor is immutable. There are _no_ methods which can validly change an existing color.

Also I think the glkVector4 change is just a sticking plaster - the proper thing to do is the final kill of all the ccColor4F junk to replace everything with CCColor with a robust set of accessors and converters where needed.

The most annoying thing is this line in CCNode init : 
 _displayColor = _color = [CCColor whiteColor].ccColor4f;

so it all starts of as a CCColor, gets downgraded to a ccColor4F and then upconverted when some real work needs doing..... dontcha just love legacy code? ;-)

pkclsoft:
@richardgrives; it's conceivable that by calling one of the initXXX methods a second time you can bypass the immutable nature of CCColor and in so doing, leak another CGColor.  i was just trying to block any possible path to another leak.

Yes the glkVector4 is a bit that way but unless you have another way to eliminate these senseless object creations for no real gain (they take time and memory at runtime) then I think this solution is quite effective.

Whilst I can see there is a real OO advantage to these accessors and converters, they are not very efficient, especially when we have code that is creating these objects ever frame update.

What I find more annoying is the code in say CCNode.colorRGBA which creates a copy to be returned rather than just returning the instance that is the actual property.  The only reason I didn't just change that is that I don't know the potential impact/risk.

richardgroves:
Calling initXXX a second (or more) times on an object is undefined behaviour (http://stackoverflow.com/questions/1989532/calling-init-multiple-times-in-objective-c) , and thus anyone doing it gets whatever is coming (even if one answer in that SO argues for defensive coding). The point with CCColor is that they are _supposed_ to be immutable - it can allow a bunch of efficiencies and optimisations. [Apple doc ref: https://developer.apple.com/library/mac/documentation/General/Conceptual/CocoaEncyclopedia/Initialization/Initialization.html - "Once an object is initialized, you should not initialize it again." ] 

On the allocation etc. during CCNode.colorRGBA - yep, that looks like a legacy of the partial move to a proper color object. If it just stored the CCColor it was created/set with then it would all be fine and the glVector4 creation would be efficient, but it still just keeps the ccColor4f struct internally. Like you I'm loathe to alter something so deep in such a core class, but at some point it needs doing (v4 has removing all the other color types as a goal, but who knows where that is now - maybe a 3.5 can move the other accessors to legacy and have the core object as the CCColor).

slembcke:
@richardgroves  Exactly. It's _definitely not_ OK to call init a second time in Objective-C. Maybe if it's a class you created, but even then it's still a really bad practice. Reinitializing a NSArray for instance is going to cause serious problems. The objects were made immutable for a reason.

As for CCColor, ccColor4f, GLKVector4, etc:
For 3.0 is was discussed that we should have a single proper color type instead of a bajillion ones like ccColor4f, ccColor3b, etc. My suggestion was that we should use GLKVector4, somebody else had the suggestion to do an object. I was worried that all the CCColor object creation would be a performance problem. Even after making some micro-benchmarks for it, the performance hit was so small that it didn't make a lot of sense to argue against it even if I thought it was creating too many temporary objects. The internal implementation uses mostly GLKVector4 or ccColor4f (for legacy reasons), and the public API uses CCColor almost exclusively. In v4, the ccColor\* structs are finally fully deprecated and replaced everywhere with GLKVector4. (Yay!) Lastly, the colorRGBA property is a legacy thing, yes. V2 had separate RGB and alpha properties so that you could change the color without changing the alpha. I was pushing for a single RGBA color property, but we ended up compromising that the old properties would stay as is and a colorRGBA property would be added instead.

The CGColor leak was fixed a while ago in Git, but it has probably not gotten into a released version:
https://github.com/cocos2d/cocos2d-objc/commit/addb32507676c47cfc1624d66ec44893c9d16cbd

pkclsoft:
So do I just close this pull request then?  I can see your points and look forward to V4.  This massive allocation does make the app look ridiculously busy in Instruments though.

Birkemose:
I have been a bit busy the last couple of days, but my understanding is that it has already been fixed. I will make sure it is merged with latest release.  
So I guess it is safe to close the request, and sorry for not being aware that there already was a fix for this.

PR #1336, opened 2015-09-05
Support for implementing gestures on nodes within a CCScrollView (inc…

nader-eloshaiker:
Support for implementing gestures on nodes within a CCScrollView (including CCTableView and CCTableViewCell).

If you implement a gestures for CCTableView (such as a swipe), then it will require to take priority over the PAN gesture within the CCScrollView.


nader-eloshaiker:
I have also added the ability to create custom CCTableViewCell with backward compatibility.

PR #1337, opened 2015-09-09
Added "requires super" attribute to cleanup method

pkclsoft:
The need to call [super cleanup] needs to be forced, or at a a minimum, encouraged by the compiler so that subclasses behave as expected by the framework.

### DIFF
--- a/cocos2d/CCNode.h
+++ b/cocos2d/CCNode.h
@@ -43,8 +43,8 @@
 @class CCBAnimationManager;
 @class CCAnimationManager;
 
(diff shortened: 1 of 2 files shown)


richardgroves:
See http://forum.cocos2d-spritebuilder.org/t/cleanup-why-is-it-now-private/17937

pkclsoft:
Yes, that's me.  I thought what I was suggesting here was exactly what you talked about.  It certainly had the effect I was looking for in my own project.

richardgroves:
Yep - just wanted to link the issue to the comment thread. Do you want to try tweaking the CCNode.h documentation on sub-classing too?

pkclsoft:
Sure.

PR #1338, opened 2015-09-20
Fix xcode 7 crash fix when using CCDrawNode

lpeancovschi:
To reproduce the crash in Xcode 7 create an empty project with SpriteBuilder and add a CCDrawNode


notthatnathan:
Thanks for this. Manually copied this into my SB project that says it's using 3.4.9, though these files didn't match. Confusing, but it works now. ¯_(ツ)_/¯

PR #1339, opened 2015-09-29
Fix loading files from full paths

charliewilliams:
Support for loading files from full paths was partially broken due to some commented-out code. This introduces a fix and calls the proper method suggested by some comments in said code.


Birkemose:
There were some bug reports that might be related to this, so until this has been tested further, I have chosen to revert the commit.

PR #1341, opened 2015-10-22
Some tweaks to the cocos2d code:

pkclsoft:
These are some changes in my project clone of cocos2d 3.4.9.  It's my first step towards formalising my tvOS branch of 3.x.  I want to sync the official 3.4.9 into my app project but I wanted to preserve these changes first.  Once this is in, and I've got my app in sync with 3.4.9 I'll start adding support for tvOS like I've done for v2.2 - classic
1. CCActionEaseOut wasn't easing very well.
2. Fixed a few basic compiler warnings.
3. Use [NSAttributedString isEqualToAttributedString:] for comparison.
4. Added enhancement to CCNode so that when a nodes content size is changed, it's children are updated as well which is essential when using normalised coordinates/sizes in children.
5. Added some more photoshop blend modes.
6. Patched NSAttributedString+CCAdditions.m to prevent font overrides when we don't want them.

### DIFF
--- a/cocos2d/CCActionEase.m
+++ b/cocos2d/CCActionEase.m
@@ -176,7 +176,7 @@
 @implementation CCActionEaseOut
 -(void) update: (CCTime) t
 {
-	[_inner update: powf(t,1/_rate)];
+    [_inner update: 1.0f - powf(1.0f - t, _rate)];
 }
 @end
 
--- a/cocos2d/CCActionInterval.m
+++ b/cocos2d/CCActionInterval.m
@@ -336,7 +336,8 @@
 	{
 		while (dt > _nextDt && _total < _times)
 		{
-			[_innerAction update:1.0];
+
+			[_innerAction update:1.0f];
 			_total++;
 
 			[_innerAction stop];
@@ -345,7 +346,7 @@
 		}
 		
 		// fix for issue #1288, incorrect end value of repeat
-		if(dt >= 1.0 && _total < _times) 
+		if(dt >= 1.0f && _total < _times) 
 		{
 			_total++;
 		}
@@ -367,7 +368,7 @@
 	}
 	else
 	{
-		[_innerAction update:fmod(dt * _times, 1.0)];
+		[_innerAction update:fmodf(dt * _times,1.0f)];
 	}
 }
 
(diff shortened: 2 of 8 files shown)


Birkemose:
Looks good, but I am not sure about 4)  
While the functionality is better / correct, it might actually break some users designs as it isn't backwards compatible, and really can't be called a bug.  
I will merge it anyways, as the number of users affected probably are very low.

pkclsoft:
Thanks @Birkemose .  If you want, I'll update it next time so that it can be turned off by default.  Similar income ways to the cascade opacity mechanism.

richardgroves:
For (1) CCTime is type 'double' so the CCActionInterval changes are a retrograde step, and the CCActionEase could do with becoming:

```
[_inner update: 1.0 - pow(1.0 - t, _rate)];
```

For (4) I like the idea of a cascadeSizeChange flag to auto-update normalised children

Birkemose:
Urk yes, all time intervals are doubles. Not that it probably will ruin much now, but would be great to get it fixed and adding the cascade flag.

richardgroves:
Minor change but CCTime should probably be defined as NSTimeInterval (which is a double anyway), and appropriately on other platforms to make the best mapping in OS/platform types for future consistency.

pkclsoft:
So @richardgroves , if I revert those 'f' suffixes and change CCTime to be NSTimeInterval this is OK?  I can also fix all of the other 'f' suffixes in those two files (there is quite a mix of this usage in there).

I'll also add the new cascade property as suggested.

richardgroves:
Sorry - didn't mean to change CCTime to be NSTimeInterval in the using code - but in the core definition at ccTypes.h - which might be a bit of a major change for a small patch, and right now has no real benefits so that might be one for when the version number is shifting.

On the 'f' functions and constants - yep, just drop the 'f': it's taking me ages to get into the habit of not automatically typing 1.0f etc. now that 64 bit is widespread and CGFloat is a double half the time....

pkclsoft:
Actually the change of CCTime in ccTypes.h is I think quite painless.  That's what I had thought you meant anyways.  I've submitted a pull request for the cascade property already.

richardgroves:
I guess I am more nervous than you about header file changes! ;-)

Saw the cascade PR - looks good to me.

Birkemose:
Cascade looks great : )

Hmmm ... I dont think I would redefine ccTime. Not that I can think of anything that would break, but you never know. If we just clean up the float / double mismatch, I think that should do it.

PR #1346, opened 2015-11-02
Added cascadeContentSizeEnabled property

pkclsoft:
Added cascadeContentSizeEnabled property so that the original, default behaviour is left intact.  Setting this new property to YES will cause contentSize changes to propagate down.


Birkemose:
Urk. I got a couple of bug reports that might relate to this, so I had to revert the commit until somebody have more time to test it.

pkclsoft:
Sorry about that.  Do you want me to look at them?  Were they with the new property turned off (the default)?

PR #1347, opened 2015-11-02
Changed a  number of 32-bit float literal constants to doubles.

pkclsoft:
Changed a  number of 32-bit float literal constants to doubles by removing the 'f' suffix.


richardgroves:
Nice work - improves it for the modern platforms.

I'm beginning to bore myself on this - but  CCActionEaseRate and CCActionEaseElastic should probably have their 'rate' and 'period' values as CGFloat's.

Line 39 of CCActionEase still leaves the (float) cast in for M_PI_X_2

pkclsoft:
Darn.  Getting bored too.

pkclsoft:
Have updated with your suggestions.

pkclsoft:
Added a new commit to fix an issue in CCScrollView on iOS 9.

PR #1356, opened 2016-01-13
FIX CCDirector's popToRootSceneWithTransition (running scene onExitTransitionDidStart and onExit where called twice)

finthamoussu:


s1ddok:
Can you reissue PR to develop-v3.5?

PR #1362, opened 2016-03-04
Fixed contentScale logic while parsing the tile map file

xmkevinchen:


s1ddok:
Can you please explain more on why this should be merged and what exactly problems does it solve?

xmkevinchen:
Sure, 

Here's the reason I started investigating the TileMaps code, I was trying to load a tile map size (1024 x 1024) on iPad Air 2, and I got the contentSize after parsing from tmx was (256 x 256). But it's supposed to be 512 x 512.  

So I started reading the code and the I found 
1.  `CCTMXXMLParser` when reads the tileSize from tmx, the code is doing the calculation based on the contentScale. But in `-[CCTMXLayer initWithTilesetInfo:layerInfo:mapInfo]` for setting the contentSize, we use the contentScale again to calculate pixelsToPoints which was no sense at this moment. The calculation has been made while parsing the tileSize.
2. Most of values for the width / height, position (x/y), offset (x/y) are presented in pixels in TMX file, except the mapSize which is kind of for  the value of row / column of tiles. So I went through the parsing method to update the calculation.
3. While debugging, I found for offset, instead of using x/y as keys, Tiled uses offsetx / offsety to present the offset, so I made the change as well

s1ddok:
I can't tell if this a correct PR or no. You could get wrong size because you use 4.0 content scale for iPad.  That is why you get 256 result pixels when loading 1024. I need some test cases to reproduce this.

s1ddok:
This PR seems to be fixing the problem with offset that was reported in the forum. Merged.

PR #1367, opened 2016-03-08
Enable cocos2d v3.4.9 to compile for mac, for SB usage

oxeron:
The CCDirector.m bug was corrected in v3.5 branch but not the old 3.4.9
Can you add this patch so that SD 1.5 can use the official cocos repository ?


s1ddok:
This is fixed in 3.5.

PR #1370, opened 2016-03-10
Header modifications for -ext

oxeron:
Modifications for using -ext


oxeron:
I made a patch from these commit
Applied to your last version
Build ok for iOS, tvOS and Mac target
Just need to remove red file CCRenderTexture_Private.h from Xcode in cocos2d/Rendering

PR #1373, opened 2016-03-21
Fixed CCTextField textField property always being nil on mac.

twomedia:
Fixes issue #1122 when building for OSX.


twomedia:
@xin-savysoda

s1ddok:
Thanks for the contribution

PR #1382, opened 2016-06-25
Resolved compiler error when compiling a project that uses ObjCpp

logan-5:
... as well as Cocos2D-Objc.  Also added an Objective-C++ test to cocos2d-tests; one involving a basic node hierarchy, the next involving a slightly more dicey manual visiting of a node with proper transforms, without necessarily adding it as a child of another node, using CCRenderer_Private.h.

This patch does _not necessarily_ allow you to compile cocos2d in Objective-C++ mode.  It only ensures that you can use Objective-C++ in your project along with Cocos2D-Objc.

The Objective-C++ fix within CCRenderer_Private.h was suggested by @slembcke here: http://forum.spritebuilder.com/t/cocos3d-c-mm-files-compiler-issue/15570/4


s1ddok:
I think this error appears only with ObjCpp compilers.. I have to look at it more precisely a bit later.

I actually use ObjC++ now in my project with 3.5 and do not face any issues, I have to check it.

